PR #5, opened 2022-07-06
BW-1284 Component to accept workflow inputs and submit run to CBAS

salonishah11:
In `config.json`, the `cbasUrlRoot` has been updated to point to port `3000` to resolve the CORS error. A local proxy in `setupProxy.js` file is setup that will redirect any /api requests made to port `8080`. More details are in [this link](https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually).

**Note to reviewers:**
Main files to review:
- package.json
- public/config.json (Note: 
- src/libs/ajax.js
- src/pages/Main.js
- src/pages/SubmitWorkflow.js
- src/pages/WorkflowInputs.js
- src/pages/WorkflowSource.js
- src/setupProxy.js

Files with components copied from Terra UI:
- src/components/common.js
- src/libs/nav.js
- src/libs/routes.js
- src/libs/state.js

Other files are auto generated and can be ignored.

Testing path Local UI -> Local CBAS Service -> Local Cromwell works fine.

**Below are snapshots from testing Cromwell app by manually deploying helm changes:**

CBAS UI prints the response it gets from CBAS after successful submission (please ignore the `WOO!`. I had added it at some point to verify the changes were actually being deployed and have removed it now)

Screen Shot 2022-07-08 at 5 17 23 PM

CBAS service deployment receiving the request in 

Screen Shot 2022-07-08 at 5 18 54 PM

Cromwell receiving the request and workflow succeeding

Screen Shot 2022-07-08 at 5 20 52 PM

**Other paths in reverse-proxy also load fine:**
/index.html
Screen Shot 2022-07-08 at 5 23 25 PM

/cbas/status
Screen Shot 2022-07-08 at 5 24 05 PM

Closes https://broadworkbench.atlassian.net/browse/BW-1284


src/pages/SubmitWorkflow.js
```diff
@@ -0,0 +1,35 @@
...
+  const [showInputsPage, setShowInputsPage] = useState(false)
+
+  return div([
+    headerBar(),
+    div({ style: { margin: '4rem' } }, [
+      h2(['Submit a workflow']),
+      div(['Submit your Terra workflows with the Cromwell engine. Full featured workflow submissions coming soon!']),
+      div({ style: { marginTop: '2rem' } }, [
+        !showInputsPage && h(WorkflowSource, { workflowUrl, setWorkflowUrl, onClick: () => setShowInputsPage(true) }),
+        showInputsPage && h(WorkflowInputs, { workflowUrl, workflowInputs, setWorkflowInputs, displayWorkflowUrlPage: () => setShowInputsPage(false) })
```
cjllanwarne, line 21: TOL: It feels to me like the `showInputsPage` decision (and therefore any buttons to change its value?) really belong at the "SubmitWorkflow" component level rather than being passed down as pre-staged functions to sub-components. (eg maybe we could pass the functions to fetch and update `workflowUrl` to the `WorkflowSource` page, but the button which progresses *this* page from its "enter URL" to its "enter inputs" state could be in this component?) What do you think?

src/pages/WorkflowInputs.js
```diff
@@ -0,0 +1,53 @@
...
+          'Outputs will be saved to cloud storage',
+          div([
+            h(ButtonOutline, {
+              onClick: displayWorkflowUrlPage
+            }, ['Back to previous page']),
+            h(ButtonPrimary, {
+              style: { marginLeft: '1rem' },
+              disabled: !workflowInputs,
+              onClick: () => submitRun()
+            }, ['Run workflow'])
```
cjllanwarne, line 47: I think this button should be at the `SubmitWorkflow` component level

src/pages/WorkflowSource.js
```diff
@@ -1,52 +1,37 @@
...
-          h(ButtonPrimary, {
-            disabled: !workflowUrl
-            // onClick: navigateToNextPage TODO: Navigate to inputs page in https://broadworkbench.atlassian.net/browse/BW-1284
-          }, ['Use workflow'])
-        ])
+    h3(['Provide a URL to a workflow']),
+    div([
+      h(Fragment, [
+        `You can browse for workflows in `,
+        makeInlineDockLink('https://firecloud.dsde-dev.broadinstitute.org/?return=terra#methods', 'Broad Methods Repository', 14),
```
cjllanwarne, line 14: `?return=terra#methods` looks like copy/paste that won't actually work in this UI?

src/setupProxy.js
```diff
@@ -0,0 +1,12 @@
...
+
+module.exports = function(app) {
+  app.use(
+    '/api',
+    createProxyMiddleware({
+      target: 'http://localhost:8080', // CBAS Service
+      changeOrigin: true
+    })
+  )
+}
```
cjllanwarne, line 12: Was this for the in-UI routing thing? I would guess we don't need this any more?

Dockerfile
```diff
@@ -3,7 +3,7 @@ FROM node:14
 RUN set -x \
   && git clone https://github.com/DataBiosphere/terra-batch-analysis-ui.git \
   && cd terra-batch-analysis-ui \
-  && git checkout main \
+  && git checkout ss_submit_workflow \
```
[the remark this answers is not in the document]
salonishah11: Since this same dockerfile is used for building images that are published to GCR, I had kept it to be checked out from `main`. Plus I am thinking that once [BW-1297](https://broadworkbench.atlassian.net/browse/BW-1297) is done, we might no longer need this for creating images and pushing it to GCR and we can update this file for locally building images to test out the UI.

public/config.json
```diff
@@ -1,3 +1,3 @@
 {
-  "cbasUrlRoot" : "http://localhost:8080"
+  "cbasUrlRoot" : "http://localhost:3000"
```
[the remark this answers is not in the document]
salonishah11: No, this  in combination with `src/setupProxy.js` was actually changed to work around the CORS error that occurs when running the UI and CBAS Service locally. UI will make the api requests to port `3000` but the local proxy in `setupProxy.js` will redirect it to port `8080`. 
Reference - https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually

src/pages/Main.js
```diff
@@ -1,26 +1,13 @@
...
+
 import { h } from 'react-hyperscript-helpers'
-import { Ajax } from 'src/libs/ajax'
-import { useCancellation, useOnMount } from 'src/libs/react-utils'
-import { WorkflowSource } from 'src/pages/WorkflowSource'
+import { LocationProvider, Router } from 'src/libs/nav'
 
 
 const Main = () => {
-  const [cbasStatus, setCbasStatus] = useState()
```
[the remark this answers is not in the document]
salonishah11: Yeah we can keep it. But we might need to talk with Tiffany to figure out the right UX for this. Maybe we can add it to bottom of the page instead of at top?

src/pages/WorkflowSource.js
salonishah11: Oops. I forgot to remove that. I will update it.

src/setupProxy.js
salonishah11: We still need this to resolve the CORS error that we get when running local UI with local CBAS Service

src/pages/SubmitWorkflow.js
salonishah11: To confirm my understanding, instead of having the child components directly update `workflowUrl` using `setWorkflowUrl`, we should pass functions to child component that are defined in parent component, and they are called by the child component? Here I was following the pattern similar to Terra UI's [Upload.js component](https://github.com/DataBiosphere/terra-ui/blob/69c85b640ec1a4a41f5388e6c7155ae35e516464/src/pages/Upload.js#L729-L918), where the child components are passed [state setters](https://github.com/DataBiosphere/terra-ui/blob/69c85b640ec1a4a41f5388e6c7155ae35e516464/src/pages/Upload.js#L873-L880) as well as [pre-staged functions](https://github.com/DataBiosphere/terra-ui/blob/69c85b640ec1a4a41f5388e6c7155ae35e516464/src/pages/Upload.js#L803)

Dockerfile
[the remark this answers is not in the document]
cjllanwarne: Definitely a topic for another ticket 😄 ... I would expect the "push to GCR" action to look kind of like:

1. Check out main
2. Build Dockerfile on what you just checked-out
3. Push to GCR

Which I don't think would be inconsistent with a local build of

1. Check out (or update) any branch
2. Build Dockerfile on what you just checked-out
3. Push to GCR

src/setupProxy.js
cjllanwarne: I remember running into this and fixing it with an annotation in CBAS itself (which I think is where the error was coming from) along the lines of "allow requests from localhost". I'm not sure which is neater, but in either case I think it's probably important not to leave it activated accidentally in production

salonishah11: Yes I had tried adding `@CrossOrigin` annotation to the /POST endpoint method handler in CBAS, but the guidance by App Sec (see [this Slack thread](https://broadinstitute.slack.com/archives/C2B7JFJ4S/p1657053506352529?thread_ts=1657036114.427909&cid=C2B7JFJ4S)) was not to have this annotation checked in as it can introduce security vulnerabilities. Instead they had suggested adding the local proxy which is only activated in development mode. I think the article also mentions that this is activated in development only.

src/pages/SubmitWorkflow.js
salonishah11: As discussed, I have moved the buttons to the parent component `SubmitWorkflow`. Now `SubmitWorkflow` will have all the orchestration related code.

aednichols (commented):
I can see why we would check in `yarn.lock` to pin specific versions, but the ZIP files in `.yarn/cache` don't seem like a fit for source control.

aednichols (approved):
Actually, that [does seem to be conventional](https://devcenter.heroku.com/articles/migrating-to-yarn-2#make-changes-to-source-code).

> The contents of `.yarn/cache` are not comparable to `node_modules`. The files are compressed and are meant to be checked into the project.

src/pages/SubmitWorkflow.js
```diff
@@ -11,14 +13,37 @@ export const SubmitWorkflow = () => {
   const [workflowInputs, setWorkflowInputs] = useState()
   const [showInputsPage, setShowInputsPage] = useState(false)
 
+  const signal = useCancellation()
+
+  const submitRun = async () => {
+    const runRes = await Ajax(signal).Cbas.submitRun(workflowUrl, workflowInputs)
+    console.log(runRes)
```
cjllanwarne, line 20: I don't think we want to log this any more?

src/pages/SubmitWorkflow.js
```diff
@@ -11,14 +13,37 @@ export const SubmitWorkflow = () => {
...
   return div([
     headerBar(),
     div({ style: { margin: '4rem' } }, [
       h2(['Submit a workflow']),
       div(['Submit your Terra workflows with the Cromwell engine. Full featured workflow submissions coming soon!']),
       div({ style: { marginTop: '2rem' } }, [
-        !showInputsPage && h(WorkflowSource, { workflowUrl, setWorkflowUrl, onClick: () => setShowInputsPage(true) }),
-        showInputsPage && h(WorkflowInputs, { workflowUrl, workflowInputs, setWorkflowInputs, displayWorkflowUrlPage: () => setShowInputsPage(false) })
+        !showInputsPage && h(WorkflowSource, { workflowUrl, setWorkflowUrl }),
+        !showInputsPage && div({ style: { display: 'flex', justifyContent: 'flex-end', marginTop: '1rem' } }, [
```
cjllanwarne, line 30: I have a suspicion you can group these together in an array, eg `!showInputsPage && [ h(...), div(...) ]`

src/pages/SubmitWorkflow.js
```diff
@@ -11,14 +13,37 @@ export const SubmitWorkflow = () => {
...
+          h(ButtonPrimary, {
+            disabled: !workflowUrl,
+            onClick: () => setShowInputsPage(true)
+          }, ['Use workflow'])
+        ]),
+        showInputsPage && h(WorkflowInputs, { workflowUrl, workflowInputs, setWorkflowInputs }),
+        showInputsPage && div({ style: { display: 'flex', marginTop: '1rem', justifyContent: 'flex-end' } }, [
+          h(ButtonOutline, {
+            onClick: () => setShowInputsPage(false)
+          }, ['Back to previous page']),
```
cjllanwarne, line 40: Maybe `"Change selected workflow"`? Since it isn't really changing what page we're on?

src/pages/WorkflowInputs.js
```diff
@@ -34,19 +25,7 @@ export const WorkflowInputs = ({ workflowUrl, workflowInputs, setWorkflowInputs,
...
-              onClick: displayWorkflowUrlPage
-            }, ['Back to previous page']),
-            h(ButtonPrimary, {
-              style: { marginLeft: '1rem' },
-              disabled: !workflowInputs,
-              onClick: () => submitRun()
-            }, ['Run workflow'])
-          ])
-        ])
+        div({ style: { marginTop: '1rem' } }, ['Outputs will be saved to cloud storage'])
```
cjllanwarne, line 28: I would put this line in the `SubmitWorkflow` component too (it's more to do with submitting the workflow than choosing its inputs)

src/pages/WorkflowSource.js
```diff
@@ -1,17 +1,17 @@
...
 
-export const WorkflowSource = ({ workflowUrl, setWorkflowUrl, onClick }) => {
+export const WorkflowSource = ({ workflowUrl, setWorkflowUrl }) => {
   return div([
     h3(['Provide a URL to a workflow']),
     div([
       h(Fragment, [
         `You can browse for workflows in `,
-        makeInlineDockLink('https://firecloud.dsde-dev.broadinstitute.org/?return=terra#methods', 'Broad Methods Repository', 14),
+        makeInlineDockLink('https://firecloud.dsde-dev.broadinstitute.org/?return=#methods', 'Broad Methods Repository', 14),
```
cjllanwarne, line 14: I actually think we don't really want a return link at all? Because presumably it will return us to... a "singleton Cromwell" Terra page that's totally unrelated to the CaaA pages we linked from?

cjllanwarne (approved):
LGTM, just a few little pieces of fixup

src/pages/SubmitWorkflow.js
salonishah11: Today I re-learned about `Fragments` that can be used to group different components/tags. Simply using `[..]` throws error that unique key needs to be assigned to each element in array, but there is no need for that in `Fragment` hence updated it 👍

src/pages/WorkflowInputs.js
salonishah11: Hmmm, I see what you mean. I will move it to `SubmitWorkflow`

src/pages/WorkflowSource.js
salonishah11: So for some reason, without `return=#methods`, the URL redirects to Terra dev instead of loading Methods UI. Same thing for `?return=`. Hence I left it like this.

kpierre13 (approved):

src/pages/Main.js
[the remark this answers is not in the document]
cjllanwarne: This is mainly a debug/validation component while we build up the app and introduce error handling... So I don't think it's worth a design. My preference would be something like:
- Have a slightly ugly element until we're confident everything is working nicely
- Implement nice "oops, Cromwell/CBAS seems to be missing" error messages to catch the cases when things are broken
- Remove the ugly element

PR #7, opened 2022-07-15
BW-1306 Add static "Saved Workflows" section

salonishah11:
This PR only adds static "Saved Workflows" section. The data displayed in table is hardcoded and will be removed in [BW-1318](https://broadworkbench.atlassian.net/browse/BW-1318). Both data columns are sortable. Clicking on `Use Workflow` button selects that particular workflow and navigates the UI to `WorkflowInputs` component.

Note for reviewers:
- Anything in `.yarn/..`, `.pnp.cjs`and `yarn.lock` is auto-generated
- content in `table.js` and `/src/libs/util.js` has been copied from Terra UI

Local testing screenshots:
When there are no saved workflows, `Nothing here yet! Your previously run workflows will be saved here.` is displayed
Screen Shot 2022-07-15 at 9 12 12 AM

When there are saved workflows (the table has vertical scrollability)
Screen Shot 2022-07-15 at 9 13 09 AM

Closes https://broadworkbench.atlassian.net/browse/BW-1306


src/pages/SavedWorkflows.js
```diff
@@ -0,0 +1,101 @@
...
+    {
+      workflowUrl: 'https://dockstore/abc/hello.wdl',
+      submissionTimestamp: '2019-07-12T22:24:15.591Z'
+    }
+  ]
+
+  const sortedPreviousRuns = _.orderBy(sort.field, sort.direction, previousRuns)
+
+  return div({ style: { marginTop: '2em', display: 'flex', flexDirection: 'column' } }, [
+    h3(['Saved Workflows']),
```
fredlas, line 57: How would you feel about changing this to "Use a saved workflow"? I'm thinking about having it match the style of the existing "Provide a URL to a workflow", rather than one being an action phrase and the other being a descriptive title.

fredlas (approved):
Looks good. I like the table entries having the same "Use Workflow" button as the existing text field - makes it clear at a glance that they're going to be putting you into the same process.

Along those lines, to reinforce the "doing the same thing, just with a new vs existing workflow" concept, what do you think about adding the word 'new' to the existing heading "Provide a URL to a [new] workflow"?

However, I don't have enough context to have any confidence in any suggestion I make here, so please feel free to go ahead without adopting them!

src/pages/SavedWorkflows.js
salonishah11: I think what you are suggesting makes sense. But I slightly prefer "Saved Workflows" as this is a whole section where (I think, and it might change later) the user would be able to see both the workflows and inputs they submitted. So this will become a place also to view those results as well as reuse both. 

Re: adding `new` in "Provide a URL to a [new] workflow" - I am thinking this might confuse users that they can't reuse previously submitted workflows? But I could be wrong. 

I feel like leaving it as it for now. Although it might be worth bringing this to UX.

PR #9, opened 2022-07-21
BW-1307 Add static "Previous Runs" page

salonishah11:
Note to reviewers:
- Files in `.yarn/..`, `pnp.cjs` and `yarn.lock` are auto generated and can be ignored
- Content in `src/components/Modal.js` and `src/components/table.js` is copied from Terra UI

Testing done locally:
Previous runs page with 3 workflow runs:

Previous runs page with 18 workflow runs with 100 page size: 

Previous runs page with 18 workflow runs with 10 page size:
page 1

page 2

Previous runs page with no workflows:

Clicking on "View" link opens a modal with expected inputs JSON:

Sorting works as expected in all cases. The page loads at path `/#previous-runs`.

Closes https://broadworkbench.atlassian.net/browse/BW-1307


kpierre13 (approved):

src/pages/PreviousRuns.js
```diff
@@ -1,8 +1,244 @@
...
+  const firstPageIndex = (pageNumber - 1) * itemsPerPage
+  const lastPageIndex = firstPageIndex + itemsPerPage
+  const paginatedPreviousRuns = sortedPreviousRuns.slice(firstPageIndex, lastPageIndex)
+
+
+  return h(Fragment, [
+    headerBar(),
+    div({ style: { margin: '4em' } }, [
+      h2(['Previous Runs']),
+      div(['Remember to turn off your Cromwell App in Terra once you are done to prevent incurring costs.']),
```
cjllanwarne, line 160: Can this be standardized or extracted? It feels like it could become quickly tedious to copy/paste it, and even more tedious to keep it all updated if it ever changes?

src/pages/PreviousRuns.js
```diff
@@ -1,8 +1,244 @@
...
+          filteredDataLength: sortedPreviousRuns.length,
+          unfilteredDataLength: sortedPreviousRuns.length,
+          pageNumber,
+          setPageNumber,
+          itemsPerPage,
+          setItemsPerPage: v => {
+            setPageNumber(1)
+            setItemsPerPage(v)
+          },
+          itemsPerPageOptions: [10, 25, 50, 100, 250, 500, 1000]
```
cjllanwarne, line 219: Maybe cap this at... 100?

src/pages/PreviousRuns.js
```diff
@@ -1,8 +1,244 @@
...
+                headerRenderer: () => h(Sortable, { sort, field: 'submissionTimestamp', onSort: setSort }, ['Submitted']),
+                cellRenderer: ({ rowIndex }) => {
+                  return h(TextCell, [Utils.makeCompleteDate(paginatedPreviousRuns[rowIndex].submissionTimestamp)])
+                }
+              }
+            ]
+          })
+        ])
+      ]),
+      !_.isEmpty(sortedPreviousRuns) && div({ style: { bottom: 0, position: 'absolute', marginBottom: '1.5rem', right: '4rem' } }, [
```
cjllanwarne, line 208: TOL: I wonder the `isEmpty` could be replaced by something like "if length is less than 10", since there's no point showing pagination options if there are not enough rows for a second page? Maybe that's an enhancement too far and would actually be confusing?

src/pages/PreviousRuns.js
```diff
@@ -1,8 +1,244 @@
...
+      width: 600,
+      onDismiss: () => setViewInputsId(undefined),
+      showCancel: false,
+      okButton:
+        h(ButtonPrimary, {
+          disabled: false,
+          onClick: () => setViewInputsId(undefined)
+        }, ['OK'])
+    }, [
+      h(TextArea, {
```
cjllanwarne, line 234: If "raw json" is likely to be the way this information is presented for at least a little while, I'd say let's try to do our best to make it as readable as possible.

There's a cool json formatter in Terra UI that might be worth a look (see an example usage in the [Workflow Dashboard](https://github.com/DataBiosphere/terra-ui/blob/dev/src/pages/workspaces/workspace/jobHistory/WorkflowDashboard.js#L197))

cjllanwarne (commented):
Tested by running locally. A few minor comments and a few observations to think about:

1. It felt a little odd to me that only the previous runs page has the warning about turning off Cromwell
2. It felt a little odd to me that the header always says "SUBMIT WORKFLOWS WITH CROMWELL", even while on the previous runs page
  - Actually, should the text just be "Batch Analysis App", or similar?

src/pages/PreviousRuns.js
salonishah11: I think it might be cleaner to show the pagination even if items are less than 10. It seems more consistent instead of having pagination appear and disappear based on length of previous runs.

src/pages/PreviousRuns.js
salonishah11: That is a good point. I will try adding that 👍

src/pages/PreviousRuns.js
salonishah11: Actually, I am thinking that does it make sense to have it here? Maybe it should be always at the footer of the page as remainder or it could be added to the Create Cloud Environments component in Terra UI? What do you think?

salonishah11:
> Actually, should the text just be "Batch Analysis App", or similar?

@cjllanwarne - that is a good point. I am inclined to set it to "Batch Analysis" as additionally down the line CBAS might submit workflows to some other engine and "Submit workflows with Cromwell" will no longer be true. It is better for CBAS UI to be backend agnostic.  But I will confirm this title change with Tiffany before merging the PR.

src/pages/PreviousRuns.js
salonishah11: I have updated it to use `ReactJson`. Thank you for the suggestion!

Screen Shot 2022-07-25 at 7 23 15 PM

src/pages/PreviousRuns.js
salonishah11: Checked it with Tiffany. There is a plan to add this warning in the Create Cloud Environments component in Terra UI. She was also going to look into adding this warning as a footer in the app. If that is the case, we might have a ticket later on to add this message to the footer and it can be removed from here. But I will leave this here for now.

src/pages/PreviousRuns.js
```diff
@@ -231,11 +231,14 @@ export const PreviousRuns = () => {
...
-      h(TextArea, {
-        style: { height: 100 },
-        'aria-label': 'Inputs JSON',
-        value: _.isEmpty(paginatedPreviousRuns[viewInputsId].workflowInputs) ? '{}' : paginatedPreviousRuns[viewInputsId].workflowInputs,
-        readOnly: true
+      h(ReactJson, {
+        style: { whiteSpace: 'pre-wrap', wordBreak: 'break-word' },
+        name: false,
+        collapsed: 4,
+        enableClipboard: false,
```
cjllanwarne, line 238: I think allowing copy-to-clipboard might be a really nice bonus, actually

salonishah11:
@cjllanwarne re: renaming title to "Batch Analysis" - It seems the direction from UX right now is to keep everything same as the mockups as it is going to be changed soon. Hence I have reverted that change and the title is "Submit workflows with Cromwell" for now.

cjllanwarne (approved):

PR #16, opened 2022-08-17
BW-1342 Support for passing entity field and ID to CBAS

salonishah11:
Note to reviewers: `TextInput` in `input.js` is copied from Terra UI

Testing: Local against local CBAS with Chris' changes from [PR #33](https://github.com/DataBiosphere/cbas/pull/33)

Screenshot of what the new inputs page looks like. The placeholders are added to let user know what is the expected structure of input:

After filling up the fields:
Screen Shot 2022-08-17 at 11 47 02 AM

Successful response from CBAS after submitting the workflow:
Screen Shot 2022-08-17 at 11 47 59 AM

Closes https://broadworkbench.atlassian.net/browse/BW-1342

### DIFF
--- a/src/pages/SubmitWorkflow.js
+++ b/src/pages/SubmitWorkflow.js
@@ -13,7 +13,9 @@ import { WorkflowSource } from 'src/pages/WorkflowSource'
 export const SubmitWorkflow = () => {
   // State
   const [workflowUrl, setWorkflowUrl] = useState()
-  const [workflowInputs, setWorkflowInputs] = useState()
+  const [workflowInputsDefinition, setWorkflowInputsDefinition] = useState()
+  const [entityType, setEntityType] = useState('')
+  const [entityId, setEntityId] = useState('')
   const [showInputsPage, setShowInputsPage] = useState(false)
   const [cbasStatus, setCbasStatus] = useState()
   const [runsData, setRunsData] = useState()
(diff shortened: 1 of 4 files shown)


kpierre13 (approved):

src/pages/SubmitWorkflow.js
```diff
@@ -41,7 +43,16 @@ export const SubmitWorkflow = () => {
 
   const submitRun = async () => {
     try {
-      await Ajax(signal).Cbas.runs.post(workflowUrl, workflowInputs)
+      const runSetsPayload = {
+        workflow_url: workflowUrl,
+        workflow_param_definitions: JSON.parse(workflowInputsDefinition),
+        wds_entities: {
+          entity_type: entityType,
+          entity_ids: JSON.parse(entityId)
```
cjllanwarne, line 51: Curious why the entityId needs parsing first?

src/pages/WorkflowInputs.js
```diff
@@ -1,28 +1,74 @@
...
-import { TextArea } from 'src/components/input'
+import { TextArea, TextInput } from 'src/components/input'
 import { FormLabel } from 'src/libs/form'
 
 
-export const WorkflowInputs = ({ workflowUrl, workflowInputs, setWorkflowInputs }) => {
+export const WorkflowInputs = ({ workflowUrl, entityType, setEntityType, entityId, setEntityId, workflowInputsDefinition, setWorkflowInputsDefinition }) => {
+  // used as placeholder to let users know expected structure of inputs definition.
+  // To be removed later when we design UI for entering the input definition.
+  const inputMappingExample = '\n[\n' +
```
cjllanwarne, line 11: Would it be easier to write this out in JSON and then use a `toString` method to generate the text box contents?

cjllanwarne (approved):
LGTM, with a few minor questions/suggestions

src/pages/SubmitWorkflow.js
salonishah11: Without the parsing, the array was passed as an escaped String and CBAS would return 400 Bad Request

src/pages/WorkflowInputs.js
salonishah11: Good point. I will try that.

PR #20, opened 2022-09-14
BW-1292 Run linting on PRs

aednichols:


src/libs/_tests/utils.test.js
```diff
@@ -1,6 +1,8 @@
 import '@testing-library/jest-dom'
+
 import { makeCompleteDate } from 'src/libs/utils'
 
+
```
aednichols, line 5: Deliberate, required to pass linter.

.github/workflows/setup-and-test.yml
```diff
@@ -27,6 +27,9 @@ jobs:
       - name: Run tests
         run: CI=true yarn test
 
+      - name: Run linter
```
salonishah11, line 30: Would it make sense to run linter before running tests?

aednichols: That's a great point, they both have the potential to fail the build but linting should be much faster, resulting in a lower average time to result

salonishah11 (approved):

.github/workflows/setup-and-test.yml
cjllanwarne: If possible in parallel would be great (ie get both results at the same time from the same PR, and not have to bounce backwards and forwards between code fixes and lint fixes)

cjllanwarne (approved):

.github/workflows/setup-and-test.yml
aednichols: The obvious way to run in parallel would be to create a new YAML and run two workflows, I'm not sure the current 5 second lint warrants that.

Ideally developers should never find out about linting problems from CI alone because local tools tell them in real time. (Any dev workflow that requires pushing to a remote to try changes gets a 👎 from me.)

aednichols: I changed my mind. The real benefit from multiple workflows is easily diagnosing whether it was linting or tests that failed, without having to click into the log.

aednichols:
Requesting re-review since the changes since last review are as big as the entire PR 😅

.github/workflows/setup-and-test.yml
```diff
@@ -24,8 +24,8 @@ jobs:
           yarn set version 3.x
           yarn install
 
-      - name: Run tests
-        run: CI=true yarn test
+      - name: Run linter
+        run: yarn eslint --max-warnings=0 .
```
salonishah11, line 28: Should this be reverted back to running tests?

aednichols: 🤦‍♂️ thank you

aednichols: And _that's_ why I re-request reviews 😄

salonishah11 (approved):
💯

PR #21, opened 2022-09-20
[BW-1376] Add workflow url and workflow inputs back

salonishah11:
Screenshots from local testing:

Since CBAS doesn't store workflow names yet, the decision made was to show Run ID for now ([see Slack thread](https://broadinstitute.slack.com/archives/G01CVT4GJ8H/p1663346166577329))

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -56,13 +56,6 @@ const Cbas = signal => ({
     get: async () => {
       const res = await fetchCbas(`runs`, { signal, method: 'GET' })
       return res.json()
-    },
-    post: async (workflowUrl, inputs) => {
-      const formData = new FormData()
-      formData.set('workflow_url', workflowUrl)
-      formData.set('workflow_params', inputs)
-      const res = await fetchCbas(`runs`, { body: formData, signal, method: 'POST' })
-      return res.json()
     }
   },
   runSets: {
--- a/src/pages/PreviousRuns.js
+++ b/src/pages/PreviousRuns.js
@@ -1,8 +1,10 @@
 import _ from 'lodash/fp'
 import { Fragment, useState } from 'react'
 import { div, h, h2 } from 'react-hyperscript-helpers'
+import ReactJson from 'react-json-view'
 import { AutoSizer } from 'react-virtualized'
-import { ButtonOutline, headerBar } from 'src/components/common'
+import { ButtonOutline, ButtonPrimary, headerBar, Link } from 'src/components/common'
+import Modal from 'src/components/Modal'
 import { FlexTable, paginator, Sortable, tableHeight, TextCell } from 'src/components/table'
 import { Ajax } from 'src/libs/ajax'
 import * as Nav from 'src/libs/nav'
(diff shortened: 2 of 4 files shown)


src/libs/ajax.js
```diff
@@ -56,13 +56,6 @@ const Cbas = signal => ({
     get: async () => {
       const res = await fetchCbas(`runs`, { signal, method: 'GET' })
       return res.json()
-    },
-    post: async (workflowUrl, inputs) => {
-      const formData = new FormData()
```
salonishah11, line 61: I removed this since CBAS no longer supports this API.

kpierre13 (approved):

src/pages/SavedWorkflows.js
```diff
@@ -25,10 +26,10 @@ export const SavedWorkflows = ({ runsData }) => {
           columns: [
             {
               size: { basis: 500 },
-              field: 'workflowName',
-              headerRenderer: () => h(Sortable, { sort, field: 'workflowName', onSort: setSort }, ['Workflow Name']),
+              field: 'workflowUrl',
+              headerRenderer: () => h(Sortable, { sort, field: 'workflowUrl', onSort: setSort }, ['Workflow Link']),
               cellRenderer: ({ rowIndex }) => {
-                return h(TextCell, [sortedPreviousRuns[rowIndex].name])
+                return h(TextCell, [sortedPreviousRuns[rowIndex].workflow_url])
```
mspector, line 32: just a small style comment: is this in snake case on purpose, to differentiate it from other places where we use `workflowUrl`? should we use camelCase?

more broadly: have we considered auto-formatting for this repo?

src/pages/PreviousRuns.js
```diff
@@ -60,21 +63,31 @@ export const PreviousRuns = () => {
...
               {
-                size: { basis: 300 },
-                field: 'workflowName',
-                headerRenderer: () => h(Sortable, { sort, field: 'workflowName', onSort: setSort }, ['Workflow Name']),
+                size: { basis: 350 },
+                field: 'runId',
+                headerRenderer: () => h(Sortable, { sort, field: 'runId', onSort: setSort }, ['Run ID']),
                 cellRenderer: ({ rowIndex }) => {
-                  return h(TextCell, [paginatedPreviousRuns[rowIndex].name])
+                  return h(TextCell, [paginatedPreviousRuns[rowIndex].run_id])
```
mspector, line 70: same comment here about snake vs. camel case

src/pages/SavedWorkflows.js
aednichols: A common reason that may be at play here is that JSONs almost always seem to use snake case for keys, and when they are parsed into a codebase that uses camel case they look out of place.

src/pages/PreviousRuns.js
salonishah11: As Adam said, since CBAS returns `run_id` and this is extracting keys from the response JSON object, it needs to be the same. If we use `runId` here, it doesn't find any such key in the `paginatedPreviousRuns` and displays nothing on the UI.

salonishah11:
@mspector One can run `yarn lint` on command line to see linting issues. We also have Github action to run the linter on every PR.

PR #27, opened 2022-10-05
[WM-1449] Display last_modified_timestamp in Previous Runs page

salonishah11:
Screen Shot 2022-10-05 at 2 22 36 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1449

### DIFF
--- a/src/pages/PreviousRuns.test.js
+++ b/src/pages/PreviousRuns.test.js
@@ -32,6 +32,7 @@ describe('Previous Runs page', () => {
         state: 'Submitted',
         submission_date: '2022-01-27T22:27:15.591Z',
         workflow_url: 'https://abc.wdl',
+        last_modified_timestamp: '2022-01-28T22:30:05.291Z',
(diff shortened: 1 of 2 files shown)


mspector (approved):
LGTM!

src/pages/PreviousRuns.test.js
```diff
@@ -170,15 +177,17 @@ describe('Previous Runs page', () => {
 
     // Assert that sort by Status worked
     const updatedDataRow1Cells = within(rows[1]).queryAllByRole('cell')
-    expect(updatedDataRow1Cells.length).toBe(4)
+    expect(updatedDataRow1Cells.length).toBe(5)
     within(updatedDataRow1Cells[0]).getByText('b7234aae-6f43-405e-bb3a-71f924e09825')
     within(updatedDataRow1Cells[1]).getByText('Failed')
-    within(updatedDataRow1Cells[3]).getByText(/Jul 14, 2022/)
+    within(updatedDataRow1Cells[2]).getByText(/Jul 14, 2022/)
+    within(updatedDataRow1Cells[4]).getByText(/Jul 14, 2022/)
```
cjllanwarne, line 184: Do we want to assert that the time is present and correct as well as the date?

cjllanwarne (approved):

src/pages/PreviousRuns.test.js
salonishah11: Asserting the timestamps is a little complicated. Since the timestamps in the data is being converted to Local timezone, it returns different time when the tests are run locally vs in GitHub action. As a result everywhere we only compare dates.

PR #31, opened 2022-11-01
WX-824 Remove URI-viewing from dashboard

jgainerdewar:
Not in scope for this ticket, but I left the `workflowLog` reference in because I thought we might want to display that path, rather than the contents of the file, in the short term.

Do folks think it's useful to leave this stuff in, commented out, as it is here?


kraefrei (approved):
LGTM, I think there is an argument to be made to just remove the commented code since its version controlled, but in the spirit of breadcrumbs for later I think I'm good with it in this case.

aednichols (approved):

PR #32, opened 2022-11-07
WX-826 WX-827 Populate workflow dashboard with data from Cromwell

jgainerdewar:
This PR adds allows workflow-dashboard-in-CBAS to load with data from a colocated Cromwell instance. There are currently no links to or from CBAS, you need to navigate in your browser to `http://localhost:3000/#job-history/<cromwell workflow id>`. Note that the Cromwell workflow id corresponds to the CBAS engine id, not the CBAS run id.

In this PR I've explicitly referred to "Cromwell" and "workflow id," we may want to make this language more generic as part of a larger workflow-dashboard-in-CBAS language revision.


src/components/job-common.js
```diff
@@ -124,30 +122,3 @@ export const makeSection = (label, children, { style = {} } = {}) => div({
 ])
 
 export const breadcrumbHistoryCaret = icon('angle-right', { size: 10, style: { margin: '0 0.25rem' } })
```
jgainerdewar, line 124: We plan to add CBAS-specific breadcrumbs in the future.

src/setupProxy.js
```diff
@@ -16,4 +16,14 @@ module.exports = function(app) {
...
     })
   )
+  app.use(
+    '/cromwell/api*',
+    createProxyMiddleware({
+      target: 'http://localhost:8000', // Cromwell
+      changeOrigin: true,
+      pathRewrite: {
+        '^/cromwell/api': '/api'
+      }
```
jgainerdewar, line 26: This proxy rule is only used when running locally.

src/libs/ajax.js
```diff
@@ -46,6 +47,7 @@ const withUrlPrefix = _.curry((prefix, wrappedFetch) => (path, ...args) => {
 export const fetchOk = _.flow(withInstrumentation, withCancellation, withErrorRejection)(fetch)
 
 const fetchCbas = withUrlPrefix(`${getConfig().cbasUrlRoot}/api/batch/v1/`, fetchOk)
+const fetchCromwell = withUrlPrefix(`${getConfig().cbasUrlRoot}/cromwell/api/workflows/v1/`, fetchOk)
```
jgainerdewar, line 50: I have not tested it but I believe that structuring the path this way should make it Just Work in the app.

kraefrei (approved):
LGTM!

aednichols (approved):

PR #56, opened 2022-12-06
WM-1585 Add columns to workflow details page

kpierre13:


src/pages/SubmissionDetails.js
```diff
@@ -36,34 +39,80 @@ export const SubmissionDetails = ({ submissionId }) => {
       }
     }
 
+    // TODO: "Prework" for the next ticket
```
mspector, line 42: is the idea that this will get un-commented in the next ticket?

If so: I get it, but would still prefer not to check in this sort of thing...

src/pages/SubmissionDetails.js
```diff
@@ -36,34 +39,80 @@ export const SubmissionDetails = ({ submissionId }) => {
...
       div({
         style: {
           marginTop: '1em', height: tableHeight({ actualRows: paginatedPreviousRuns.length, maxRows: 12.5, heightPerRow: 250 }), minHeight: '10em'
         }
       }, [
+        div([h2(['Workflows'])]),
+        div([h3(['Filter by: '])]),
+        h(Select, {
+          isDisabled: false,
+          'aria-label': 'Select a data table',
```
mspector, line 110: is this label correct? I could imagine be able to filter workflows by data table, but not for it to be the only way to filter

src/pages/SubmissionDetails.js
kpierre13: It is, but I can get rid of it!

mspector (approved):
thank you!

src/pages/SubmissionDetails.js
```diff
@@ -77,48 +126,30 @@ export const SubmissionDetails = ({ submissionId }) => {
...
-                  // link to workflow-dashboard / :workflowId
-                  return h(
-                    Link,
-                    { onClick: () => { Nav.goToPath('workflow-dashboard', { workflowId: paginatedPreviousRuns[rowIndex].engine_id }) }, style: { fontWeight: 'bold' } },
-                    ['Workflow Dashboard']
-                  )
-                }
-              },
-              {
-                size: { basis: 220, grow: 0 },
```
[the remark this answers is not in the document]
kpierre13: Updated the `Run ID` column to lead to the workflow details page when clicked.

src/pages/SubmissionDetails.js
```diff
@@ -126,7 +109,7 @@ export const SubmissionDetails = ({ submissionId }) => {
               {
                 size: { basis: 350 },
                 field: 'record_id',
-                headerRenderer: () => h(Sortable, { sort, field: 'record_id', onSort: setSort }, ['Workflow name']),
+                headerRenderer: () => h(Sortable, { sort, field: 'record_id', onSort: setSort }, ['ID']),
```
cjllanwarne, line 112: Should be "<Record Type> ID" (eg "Sample ID" for samples)

cjllanwarne (approved):
LGTM, one minor fixup on the column name

PR #57, opened 2022-12-07
render data selector table, fetch from WDS

mspector:


src/components/PopupTrigger.js
```diff
@@ -0,0 +1,152 @@
+import _ from 'lodash/fp'
```
mspector, line 1: this whole file is a copy/paste job from terra-ui, kept as similar as possible without importing or defining unnecessary code

src/components/data/data-utils.js
```diff
@@ -0,0 +1,83 @@
+import _ from 'lodash/fp'
```
mspector, line 1: also a copy/paste job -- with a lot removed compared to terra-ui's version

src/components/keyboard-nav.js
```diff
@@ -0,0 +1,97 @@
+import _ from 'lodash/fp'
```
mspector, line 1: also a copy/paste job from terra-ui

src/components/table.js
```diff
@@ -355,3 +399,295 @@ FlexTable.propTypes = {
   }),
   readOnly: PropTypes.bool
 }
+
+/**
+ * A virtual table with a fixed header and explicit column widths. Intended for displaying large
+ * datasets which may require horizontal scrolling.
+ */
+export const GridTable = forwardRefWithName('GridTable', ({
```
mspector, line 407: `GridTable` is central to this PR. this object is copy/pasted from the DataTable class in terra-ui

src/libs/ajax.js
```diff
@@ -96,53 +97,25 @@ const Cromwell = signal => ({
   }
 })
 
+// TODO: how to get these values?
+const wdsInstanceId = '15f36863-30a5-4cab-91f7-52be439f1175'
+const wdsApiVersion = 'v0.2'
```
mspector, line 102: we'll need to decide how to get these in here...

src/pages/SubmissionConfig.js
```diff
@@ -26,44 +33,57 @@ export const SubmissionConfig = ({ methodId }) => {
   const [runSetName, setRunSetName] = useState()
   const [runSetDescription, setRunSetDescription] = useState()
 
+  // TODO: this should probably be moved to a scope more local to the data selector
+  const [sort, setSort] = useState({ field: 'name', direction: 'asc' })
+
   const signal = useCancellation()
 
-  useOnMount(() => {
-    const loadMethodsData = async () => {
```
mspector, line 32: these next few changes pull out the `load<X>Data` methods out of the `useOnMount` call. I did this because `loadRecordsData` needs to be outside of this scope, and thought keeping some inside and some outside of `useOnMount` signaled a distinction that doesn't really exist (basically, I think it's slightly more understandable this way)

src/pages/SubmissionConfig.js
```diff
@@ -26,8 +33,20 @@ export const SubmissionConfig = ({ methodId }) => {
   const [runSetName, setRunSetName] = useState()
   const [runSetDescription, setRunSetDescription] = useState()
 
+  // TODO: this should probably be moved to a scope more local to the data selector
+  const [sort, setSort] = useState({ field: 'name', direction: 'asc' })
+
   const signal = useCancellation()
 
+  const loadRecordsData = async recordType => {
```
mspector, line 41: IMO, all of these `load<X>Data` functions should be outside of the `useOnMount` scope. I had that change here originally, but it made the diff harder to interpret

src/libs/ajax.js
cjllanwarne: I'd suggest making a ticket. I think we'll need Leo to provide this and render it into a config file, and also do the same in CBAS

src/pages/SubmissionConfig.js
```diff
@@ -15,9 +21,10 @@ export const SubmissionConfig = ({ methodId }) => {
   const [activeTab, setActiveTab] = useState({ key: 'select-data' })
   const [dataTables, setDataTables] = useState()
   const [methodsData, setMethodsData] = useState({})
+  const [recordsData, setRecordsData] = useState({})
 
   // Options chosen on this page:
-  const [selectedTable, setSelectedTable] = useState()
+  const [selectedTableName, setSelectedTableName] = useState()
```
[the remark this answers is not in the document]
cjllanwarne: Maybe `selectedRecordType` is even clearer?

src/pages/SubmissionConfig.js
```diff
@@ -26,8 +33,20 @@ export const SubmissionConfig = ({ methodId }) => {
   const [runSetName, setRunSetName] = useState()
   const [runSetDescription, setRunSetDescription] = useState()
 
+  // TODO: this should probably be moved to a scope more local to the data selector
+  const [sort, setSort] = useState({ field: 'name', direction: 'asc' })
```
cjllanwarne, line 37: Yes, I think record selector would be a great component to break out

cjllanwarne (approved):

kpierre13 (approved):

PR #63, opened 2022-12-13
[WM-1596] Update Submission details test

kpierre13:


src/pages/SubmissionDetails.test.js
```diff
@@ -0,0 +1,229 @@
...
+  })
+
+  afterAll(() => {
+    Object.defineProperty(HTMLElement.prototype, 'offsetHeight', originalOffsetHeight)
+    Object.defineProperty(HTMLElement.prototype, 'offsetWidth', originalOffsetWidth)
+  })
+
+  it('should correctly display previous 2 runs', async () => {
+    // Act
+    await act(async () => {
```
mspector, line 119: note: there's no need to change anything here. I'm just passing on some info I learned the other day from Miguel.

apparently, it's a "best practice" not to use `act` unless you really need it. I don't know what situation calls for it or not -- if you felt like it, you could see if this test works the same way if you replace these three lines with just `render(h(SubmissionDetails))` (without the wrapping `act` and without `await`s)

mspector (approved):
looks good -- your call as to whether or not to pull out those common functions. we could talk about where it might make sense to put them

src/pages/SubmissionDetails.js
```diff
@@ -36,10 +50,35 @@ export const SubmissionDetails = ({ submissionId }) => {
         notify('error', 'Error loading previous runs', { detail: await (error instanceof Response ? error.text() : error) })
       }
     }
+
+    const loadRunSetData = async () => {
```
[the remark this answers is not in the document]
kpierre13: I think so. I think there are several functions that could be put in a common location as there used in numerous places.

src/pages/SubmissionDetails.test.js
kpierre13: Oh, thank you. I'll test this out and see if anything changes.

kpierre13: I was able to change this in one place, but it seems that data wasn't able to be rendered when omitting the `act` from tests that were testing table functions/features.

src/pages/SubmissionDetails.js
```diff
@@ -36,10 +50,35 @@ export const SubmissionDetails = ({ submissionId }) => {
...
+    }
+
     loadRunsData()
+    loadRunSetData()
+    loadMethodsData()
   })
 
-
+  const specifyRunSet = _.filter(r => r.run_set_id === submissionId, runSetData)
+  const methodId = specifyRunSet[0]?.method_id
```
cjllanwarne, line 80: Note to self: this will need fixing with my "method -> method_version" changes

cjllanwarne (approved):

PR #64, opened 2022-12-14
[WM-1597] Add dropdown options to "Filter by"

kpierre13:
Only including an option to filter by workflows with `error_messages`.


src/pages/SubmissionDetails.js
```diff
@@ -80,7 +81,12 @@ export const SubmissionDetails = ({ submissionId }) => {
   const specifyRunSet = _.filter(r => r.run_set_id === submissionId, runSetData)
   const methodId = specifyRunSet[0]?.method_id
   const getSpecificMethod = _.filter(m => m.method_id === methodId, methodsData)
-  const sortedPreviousRuns = _.orderBy(sort.field, sort.direction, runsData)
+
+  // hardcoded to return error_messages since this is the only option we currently want to filter by
```
cjllanwarne, line 85: Is this comment still right?

src/pages/SubmissionDetails.js
```diff
@@ -80,7 +81,12 @@ export const SubmissionDetails = ({ submissionId }) => {
...
+  const errorStates = ['SYSTEM_ERROR', 'EXECUTOR_ERROR']
+  const filteredPreviousRuns = filterOption ? _.filter(r => errorStates.includes(r.state), runsData) : runsData
+  const sortedPreviousRuns = _.orderBy(sort.field, sort.direction, filteredPreviousRuns)
+  const filterOptions = ['Error']
```
cjllanwarne, line 89: We probably want a "None" option so we can turn off the filter again without reloading the page?

cjllanwarne (approved):
LGTM as a first implementation. I do think we want a "None" option too though (either in this ticket or in a follow up)

src/pages/SubmissionDetails.js
kpierre13: Removed

src/pages/SubmissionDetails.js
```diff
@@ -42,6 +43,24 @@ export const SubmissionDetails = ({ submissionId }) => {
       Utils.differenceFromNowInSeconds(submitted)
   }
 
+  const getFilter = filter => {
+    let filterStatement
+    switch (filter) {
+      case 'Error':
+        filterStatement = _.filter(r => errorStates.includes(r.state), runsData)
```
mspector, line 50: I went back and forth on whether or not this is too much of a nit-pick, but in the end decided it was worth saying...

in each of these cases, you're not just defining a filter - you're actually performing the filter operation. So `getFilter` isn't quite accurate - it's really something like `getFilteredRunsData` (or something)

You could just rename the function, which would be fine. Or...

if you wanted, you could make some small adjustments to this function to make it more "functional" (as in "functional programming"): just drop the `runsData` in this line, and line 56.

Why does that work? because in `lodash/fp`, functions that are called without all of its arguments automatically return another function that will accept the missing arguments (this is called "currying").

if you do this, then this function will truly be returning a "filter" -- a function that accepts `runsData` and returns filtered data.

(lmk if you want to discuss this in more detail)

src/pages/SubmissionDetails.js
```diff
@@ -42,6 +43,24 @@ export const SubmissionDetails = ({ submissionId }) => {
...
+        break
+      case 'None':
```
mspector, line 52: if you remove this case, will it use the default case when filter is 'None'?

if switch statements work like that -- then probably best to remove this case, since it's identical to the default

src/pages/SubmissionDetails.js
mspector: * (why is all of the above a good thing? because if you get rid of `runsData`, the function becomes "pure" (i.e. it has no side effects), which is generally considered to be easier to reason about and test

src/pages/SubmissionDetails.js
```diff
@@ -80,7 +99,11 @@ export const SubmissionDetails = ({ submissionId }) => {
   const specifyRunSet = _.filter(r => r.run_set_id === submissionId, runSetData)
   const methodId = specifyRunSet[0]?.method_id
   const getSpecificMethod = _.filter(m => m.method_id === methodId, methodsData)
-  const sortedPreviousRuns = _.orderBy(sort.field, sort.direction, runsData)
+
+  const errorStates = ['SYSTEM_ERROR', 'EXECUTOR_ERROR']
+  const filteredPreviousRuns = filterOption ? getFilter(filterOption) : runsData
```
mspector, line 104: FYI: if you make `getFilter` more "functional" according to the suggestion above, this could become:

```
const filteredPreviousRuns = filterOption ? getFilter(filterOption)(runsData) : runsData
```

src/pages/SubmissionDetails.js
kpierre13: I went back and forth on this. In the list of dropdown menu options, `None` is listed, and I figured if it was passed as an input to the function, then it's fine if there isn't a case for it as it be recognized as `default`, but I wasn't sure how specific to be for the cases. I'll get rid of it however because it is returning the same data.

src/pages/SubmissionDetails.js
```diff
@@ -43,6 +43,21 @@ export const SubmissionDetails = ({ submissionId }) => {
       Utils.differenceFromNowInSeconds(submitted)
   }
 
+  const getFilter = filter => {
```
cjllanwarne, line 46: Super nit-picky: I would rename this to `applyRunsFilter` since it is actively doing the filtering rather than getting a filter to apply later.

cjllanwarne (approved):

src/pages/SubmissionDetails.js
mspector: this may not be true anymore after Katrina's changes to make this more "functional" - now it's returning a filter function

PR #65, opened 2022-12-16
[WM-1570] submission config input table

mspector:


src/pages/SubmissionConfig.js
```diff
@@ -25,19 +26,21 @@ export const SubmissionConfig = ({ methodId }) => {
 
   // Options chosen on this page:
   const [selectedRecordType, setSelectedRecordType] = useState()
-  const [selectedRecords, setSelectedRecords] = useState()
+  const [selectedRecords, setSelectedRecords] = useState({})
   const [configuredInputDefinition, setConfiguredInputDefinition] = useState()
   const [configuredOutputDefinition, setConfiguredOutputDefinition] = useState()
 
+  const [cachedInputSources, setCachedInputSources] = useState()
```
mspector, line 33: this allows the user to navigate away from the Inputs tab without losing their selections

src/pages/SubmissionConfig.js
```diff
@@ -145,8 +149,8 @@ export const SubmissionConfig = ({ methodId }) => {
         onChangeTab: v => setActiveTab({ key: v }),
         finalStep: h(ButtonPrimary, {
           style: { marginLeft: '1rem' },
-          // disabled: !!Utils.computeWorkspaceError(ws) || !!noLaunchReason || currentSnapRedacted || !!snapshotReferenceError,
-          // tooltip: Utils.computeWorkspaceError(ws) || noLaunchReason || (currentSnapRedacted && 'Workflow version was redacted.'),
+          disabled: _.isEmpty(selectedRecords),
```
mspector, line 152: I kept trying to submit my workflow without any records selected, which errored out. I figured it would be better to just have this button disabled if no records are selected.

src/pages/SubmissionConfig.js
```diff
@@ -370,6 +372,151 @@ const renderGrid = props => {
...
+        {
+          headerRenderer: () => h(Fragment, [
+            h(HeaderCell, ['Attribute'])
+          ]),
+          cellRenderer: ({ rowIndex }) => {
+            const source = _.get(`${rowIndex}.source`, configuredInputDefinition)
+            return Utils.switchCase(source.type || 'default',
+              ['record_lookup', () => recordLookupSelect(rowIndex)],
+              ['literal', () => parameterValueSelect(rowIndex)],
+              ['default', () => h(TextCell, {}, ['Use value from Workflow'])]
```
mspector, line 511: when I saw the mockup, I thought the "use default" option should figure out what the default variable would be, and then display it here - but that would require us to actually parse the WDL, which seemed out-of-scope.

I want to confirm that this should literally say "Use value from Workflow"...

src/pages/SubmissionConfig.js
```diff
@@ -370,6 +372,151 @@ const renderGrid = props => {
...
+        const newConfig = _.set(`${rowIndex}.source`, newSource, configuredInputDefinition)
+        setConfiguredInputDefinition(newConfig)
+        setCachedInputSources(_.set(`${rowIndex}.record_attribute`, newSource.record_attribute, cachedInputSources))
+      },
+      placeholder: 'Select Attribute',
+      options: _.keys(dataTableAttributes),
+      // ** https://stackoverflow.com/questions/55830799/how-to-change-zindex-in-react-select-drowpdown
+      styles: { container: old => ({ ...old, display: 'inline-block' }), menuPortal: base => ({ ...base, zIndex: 9999 }) },
+      menuPortalTarget: document.body,
+      menuPlacement: 'top'
```
mspector, line 418: this crudely solves an annoying problem: the rows at the bottom of the inputs selector table don't have enough space to accommodate the dropdown menus. I tried to make space so that the user could scroll down, but in the end everything I tried failed.

This solution looks a bit odd (have all dropdowns "drop up"), but it works... improving this should be considered a "nice to have", in my opinion

src/pages/SubmissionConfig.js
kpierre13: We opted to say "Use value..." (as opposed to "Default value...") since the other options are all verbs/calls to action as well, right?

kpierre13 (approved):

src/pages/SubmissionConfig.js
```diff
@@ -163,14 +167,12 @@ export const SubmissionConfig = ({ methodId }) => {
...
   const renderInputs = () => {
-    return configuredInputDefinition ? h(ReactJson, {
-      style: { whiteSpace: 'pre-wrap' },
-      name: false,
-      collapsed: 4,
-      enableClipboard: false,
-      displayDataTypes: false,
-      displayObjectSize: false,
-      src: configuredInputDefinition
+    return configuredInputDefinition ? h(renderInputTable, {
```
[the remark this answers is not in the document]
cjllanwarne: Incidentally, another great reason to make this a separate component. We almost certainly want to be able to show historical input and output definitions on the Submission Details page (the spiritual successor to the "Show Inputs" button, albeit at a submission level rather than in-table-per-workflow level). 

image

Having this be a component that can either be read/write for config or read-only for record keeping would be super nice.

src/pages/SubmissionConfig.js
```diff
@@ -370,6 +372,151 @@ const renderGrid = props => {
...
+        const newSource = {
+          type: _.get(`${rowIndex}.source.type`, configuredInputDefinition),
+          record_attribute: newAttribute
+        }
+        const newConfig = _.set(`${rowIndex}.source`, newSource, configuredInputDefinition)
+        setConfiguredInputDefinition(newConfig)
+        setCachedInputSources(_.set(`${rowIndex}.record_attribute`, newSource.record_attribute, cachedInputSources))
+      },
+      placeholder: 'Select Attribute',
+      options: _.keys(dataTableAttributes),
```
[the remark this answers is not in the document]
cjllanwarne: Also... not a huge fan of the field names being broken across multiple lines, if that's easy to fix?

src/pages/SubmissionConfig.js
mspector: this might have just been a dumb moment on my part - I'll either just read from the configuredInputDefinition, or I'll come back and explain why this second state is needed

src/pages/SubmissionConfig.js
mspector: > Yeah I think the "value from WDL" is going to need more thinking about (in particular, at workflow parsing time). For now I think we can get the same effect by prepopulating the WDL values into the template run set in the database, and calling it a suggested "literal" value.

this sounds like a thing to deal with in another ticket -- assuming I'm interpreting that correctly, I'll make updates based on your following comment (i.e. change the description, only allow this for optional inputs)

src/pages/SubmissionConfig.js
```diff
@@ -370,6 +372,151 @@ const renderGrid = props => {
...
+          headerRenderer: () => h(Sortable, { sort: inputTableSort, field: 'workflowVariable', onSort: setInputTableSort }, [h(HeaderCell, ['Variable'])]),
+          cellRenderer: ({ rowIndex }) => {
+            return h(TextCell, {}, [configuredInputDefinition[rowIndex].input_name])
+          }
+        },
+        {
+          size: { basis: 160, grow: 0 },
+          headerRenderer: () => h(HeaderCell, ['Type']),
+          cellRenderer: ({ rowIndex }) => {
+            return h(TextCell, {}, [parseInputType(configuredInputDefinition[rowIndex].input_type)]) // TODO: this needs to be more flexible
```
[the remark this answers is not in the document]
mspector: might be an old comment -- I think this is good enough for target workflows

src/pages/SubmissionConfig.js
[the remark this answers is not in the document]
mspector: I have made one low-effort pass at fixing this -- I'll give it another go

src/pages/SubmissionConfig.js
mspector: I was pleasantly surprised to find that this was *not* just me being dumb, but it may have been unnecessary:

These states cache the input sources, not just for each row, but *within* that row. In other words: if they specify a `literal` value for an attribute, and then they change that attribute to `record_lookup`, they can change back to `literal` and their original value will be preserved).

src/pages/SubmissionConfig.js
```diff
@@ -370,6 +372,151 @@ const renderGrid = props => {
...
+      width,
+      columns: [
+        {
+          size: { basis: 250, grow: 0 },
+          field: 'taskVariable',
+          headerRenderer: () => h(Sortable, { sort: inputTableSort, field: 'taskVariable', onSort: setInputTableSort }, [h(HeaderCell, ['Task name'])]),
+          cellRenderer: () => {
+            return h(TextCell, { style: { fontWeight: 500 } }, [method ? method.name : 'Loading...'])
+          }
+        },
```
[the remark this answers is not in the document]
mspector: something like this?

image

src/pages/SubmissionConfig.test.js
```diff
@@ -424,7 +429,7 @@ describe('SubmissionConfig inputs definition', () => {
 
     const cells = within(rows[1]).queryAllByRole('cell')
     expect(cells.length).toBe(5)
-    within(cells[1]).getByText('target_workflow_1.foo.foo_rating')
+    within(cells[1]).getByText('foo_rating_workflow_var')
```
cjllanwarne, line 432: - Should we also check that column 0 is `foo`?
- Can we also quickly test a second input without a task (eg `input_name: 'target_workflow_1.bar_rating'`?

cjllanwarne (approved):
LGTM, only minor comments

src/pages/SubmissionConfig.test.js
mspector: good ideas. I'll add these checks

PR #69, opened 2022-12-19
[WM-1580] - Submit confirmation modal

kpierre13:


src/pages/SubmissionDetails.js
```diff
@@ -78,24 +77,11 @@ export const SubmissionDetails = ({ submissionId }) => {
...
-    const loadMethodsData = async () => {
-      try {
-        const methodsResponse = await Ajax(signal).Cbas.methods.get()
-        const allMethods = methodsResponse.methods
-        setMethodsData(allMethods)
-      } catch (error) {
-        notify('error', 'Error loading methods data', { detail: await (error instanceof Response ? error.text() : error) })
-      }
-    }
-
```
cjllanwarne, line 90: I would leave this stuff in... mainly because I will be needing it in the next PR (even if it's not 100% needed right now)

src/pages/SubmissionDetails.js
```diff
@@ -126,7 +112,7 @@ export const SubmissionDetails = ({ submissionId }) => {
       ]),
       div({ style: { marginLeft: '4em' } }, [
         h(TextCell, [(h(Link, { onClick: () => Nav.goToPath('submission-history') }, ['Submission History'])), ' >', ` Submission ${submissionId}`]),
-        h2(['workflow: ', getSpecificMethod[0]?.name]),
+        h2(['workflow: ', specifyRunSet[0]?.run_set_name]),
```
cjllanwarne, line 115: Maybe move the workflow name down into the `h3` details, but we still want it. And this header should now probably be titled "Run Set: "

cjllanwarne (approved):
One change that I think ought to be undone, otherwise LGTM

src/pages/SubmissionDetails.js
```diff
@@ -124,9 +124,10 @@ export const SubmissionDetails = ({ submissionId }) => {
           onClick: () => Nav.goToPath('root')
         }, ['Submit a new workflow'])
       ]),
-      div({ style: { marginLeft: '4em' } }, [
+      div({ style: { marginLeft: '4em', lineHeight: 1.25 } }, [
```
cjllanwarne, line 127: Seems a little suspicious that you'd need to manually edit line heights here? TOL maybe one day we can extract a style out of this so that all of the pages in the UI can benefit from the same style of lists?

src/pages/SubmissionDetails.test.js
```diff
@@ -222,7 +222,8 @@ describe('Submission Details page', () => {
 
 
     await screen.getByText(/Submission e8347247-4738-4ad1-a591-56c119f93f58/)
-    await screen.getByText(/workflow: Hello world/)
+    await screen.getByText(/Submission name: hello world/)
+    await screen.getByText(/Workflow name: Hello world/)
```
cjllanwarne, line 226: Maybe we can change one of these names to prove that the UI elements are picking things out from the right parts of the API response?

cjllanwarne (approved):

mspector (approved):

PR #75, opened 2023-01-11
[WM-1615] make network calls sequentially in SubmissionDetails

mspector:
Fixes bug reported in https://broadworkbench.atlassian.net/browse/WM-1610

### DIFF
--- a/src/pages/SubmissionDetails.js
+++ b/src/pages/SubmissionDetails.js
@@ -116,9 +116,8 @@ export const SubmissionDetails = ({ submissionId }) => {
     }
 
     loadRunsData()
-    loadRunSetData().then(runSet => {
-      runSet && loadMethodsData(runSet.method_version_id)
-    })
+      .then(() => loadRunSetData())
+      .then(runSet => runSet && loadMethodsData(runSet.method_version_id))
   })
 
   const specifyRunSet = filter(r => r.run_set_id === submissionId, runSetData)


src/pages/SubmissionDetails.js
```diff
@@ -116,9 +116,8 @@ export const SubmissionDetails = ({ submissionId }) => {
     }
 
     loadRunsData()
-    loadRunSetData().then(runSet => {
-      runSet && loadMethodsData(runSet.method_version_id)
-    })
+      .then(() => loadRunSetData())
+      .then(runSet => runSet && loadMethodsData(runSet.method_version_id))
```
salonishah11, line 120: TOL: I am wondering if there is a `.catch()` that we can use here to catch the errors from the previous failed promise and display them? Or would the `try..catch` inside individual methods be enough for that?

mspector: good question... I *think* the try/catch inside individual methods are enough

kpierre13 (approved):

src/pages/SubmissionDetails.js
salonishah11: Talked face-to-face and the `try...catch` should be sufficient.

salonishah11 (approved):

src/pages/SubmissionDetails.js
```diff
@@ -116,9 +116,8 @@ export const SubmissionDetails = ({ submissionId }) => {
     }
 
     loadRunsData()
```
cjllanwarne, line 118: Assuming we don't need runs data to load run_Set_Data, I would fetch runSet data first.

Imagine you had 100,000 runs... it's probably better to show the top level run set details while the runs are still loading in rather than delay the high level details behind loading all the runs into memory

cjllanwarne (approved):
LGTM with one small thought

src/pages/SubmissionDetails.js
mspector: good thinking, thanks!

PR #78, opened 2023-01-20
[WM-1652] move function to utils; handle potentially nested array_type

mspector:


src/libs/utils.js
```diff
@@ -160,3 +160,16 @@ export const convertValue = _.curry((type, value) => {
...
+  if (_.has('primitive_type', iotype)) {
+    return iotype.primitive_type
+  }
+  if (_.has('optional_type', iotype)) {
+    return `${_.get('optional_type.primitive_type', iotype)} (optional)`
+  }
+  if (_.has('array_type', iotype)) {
+    return `Array[${renderTypeText(_.get('array_type', iotype))}]`
+  }
+  throw new Error(`iotype ${iotype} does not contain a valid type`)
```
mspector, line 174: I'm not sure that this is the right way to handle the situation where there are no matching types

src/libs/utils.test.js
```diff
@@ -10,3 +10,13 @@ describe('makeCompleteDate', () => {
     expect(makeCompleteDate('2022-01-27T22:27:15.591Z').toString()).toContain('Jan 27, 2022')
   })
 })
+
+describe('submission-common tests', () => {
+  it('variable type text is rendered properly', () => {
+    expect(renderTypeText({ type: 'primitive', primitive_type: 'File' })).toStrictEqual('File')
+    expect(renderTypeText({ type: 'optional', optional_type: { primitive_type: 'String' } })).toStrictEqual('String (optional)')
```
cjllanwarne, line 17:
```suggestion
    expect(renderTypeText({ type: 'optional', optional_type: { type: 'primitive', primitive_type: 'String' } })).toStrictEqual('String (optional)')
```

src/libs/utils.test.js
```diff
@@ -10,3 +10,13 @@ describe('makeCompleteDate', () => {
...
+    expect(renderTypeText({ type: 'array', array_type: { primitive_type: 'Int' } })).toStrictEqual('Array[Int]')
```
cjllanwarne, line 18:
```suggestion
    expect(renderTypeText({ type: 'array', array_type: { type: 'primitive', primitive_type: 'Int' } })).toStrictEqual('Array[Int]')
```

src/libs/utils.test.js
```diff
@@ -10,3 +10,13 @@ describe('makeCompleteDate', () => {
...
+    expect(renderTypeText({ type: 'array', array_type: { optional_type: { primitive_type: 'Int' } } })).toStrictEqual('Array[Int (optional)]')
```
cjllanwarne, line 19:
```suggestion
    expect(renderTypeText({ type: 'array', array_type: { type: 'optional', optional_type: { type: 'primitive', primitive_type: 'Int' } } })).toStrictEqual('Array[Int (optional)]')
```

src/libs/utils.test.js
```diff
@@ -10,3 +10,13 @@ describe('makeCompleteDate', () => {
...
+    expect(renderTypeText({ type: 'array', array_type: { array_type: { array_type: { primitive_type: 'Int' } } } })).toStrictEqual('Array[Array[Array[Int]]]')
```
cjllanwarne, line 20:
```suggestion
    expect(renderTypeText({ type: 'array', array_type: { type: 'array', array_type: { type: 'array', array_type: { type: 'primitive', primitive_type: 'Int' } } } })).toStrictEqual('Array[Array[Array[Int]]]')
```

src/libs/utils.js
cjllanwarne: So one easy heuristic... assuming we have a bug in this logic, would we prefer the page to crash, or to display some kind of "Unrecognized Type" value and either wait to be told about it, or separately set up some kind of reporting?

cjllanwarne (approved):
Approved with a few type definition corrections and a thought on error "handling"

src/libs/utils.js
salonishah11: TOL: Maybe we could say `Type not supported` until we support all types?

salonishah11 (approved):

PR #82, opened 2023-01-24
[WM-1656] Message is shown if data table is not found

kpierre13:


kpierre13:
image

src/pages/SubmissionConfig.js
```diff
@@ -47,7 +49,7 @@ export const SubmissionConfig = ({ methodId }) => {
       const searchResult = await Ajax(signal).Wds.search.post(recordType)
       setRecords(searchResult.records)
     } catch (error) {
-      notify('error', 'Error loading WDS records', { detail: await (error instanceof Response ? error.text() : error) })
+      setNoRecordData('No existing record')
```
cjllanwarne, line 52: - I think it's the record _type_ that's not found, but I don't think users think in terms of "record type" so maybe we can just call it a data table?
- To give users a clue what to do next, can we include the table we were trying to look for here?
```suggestion
      setNoRecordData(`Data Table not found: ${recordType}`)
```

src/pages/SubmissionConfig.js
```diff
@@ -39,6 +40,7 @@ export const SubmissionConfig = ({ methodId }) => {
   const [outputTableSort, setOutputTableSort] = useState({ field: 'taskVariable', direction: 'asc' })
 
   const [launching, setLaunching] = useState(undefined)
+  const [noRecordData, setNoRecordData] = useState(null)
```
cjllanwarne, line 43: I think this is technically a record type that's missing, not a single record.

cjllanwarne (commented):
One small naming nit, and a 🤞 small change to make the "what to do next" action more obvious

cjllanwarne (approved):
LGTM. If you're feeling extra diligent you could
- Update the example screenshot
- Add a test that makes sure the text comes out right when the record is missing.

src/pages/SubmissionConfig.js
```diff
@@ -125,19 +127,26 @@ export const SubmissionConfig = ({ methodId }) => {
...
+            setNoRecordTypeData(null)
+            setSelectedRecordType(value)
+            loadRecordsData(value)
+          },
+          placeholder: 'None selected',
+          styles: { container: old => ({ ...old, display: 'inline-block', width: 200 }), paddingRight: '2rem' },
+          options: _.map(t => t.name, recordTypes)
+        }),
+        noRecordTypeData && h(Fragment, [
+          a({ style: { marginLeft: '1rem', fontSize: 15, marginTop: '1rem', height: '2rem', fontWeight: 'bold' } }, [statusType.failed.icon(), ' ', noRecordTypeData])
```
salonishah11, line 146: TOL: I am wondering if it would be better to display an orange `!` icon like below one instead of red triangle one? 🤔 
Screen Shot 2023-01-24 at 6 06 15 PM

kpierre13: I considered this. I chose the red warning since this appears due to an error in fetching from the WDS database. The orange `!` just seemed like a warning rather than an actual blocker. I don't mind changing it around however.

salonishah11: I think it would be good to check with Sara or Lou.

src/pages/SubmissionConfig.js
```diff
@@ -125,19 +128,26 @@ export const SubmissionConfig = ({ methodId }) => {
...
-        placeholder: 'None selected',
-        styles: { container: old => ({ ...old, display: 'inline-block', width: 200 }) },
-        options: _.map(t => t.name, recordTypes)
-      }),
+      div({}, [
+        h(Select, {
+          isDisabled: false,
+          'aria-label': 'Select a data table',
+          isClearable: false,
+          value: selectedRecordType ? selectedRecordType : null,
```
cjllanwarne, line 136: So definitely a small complaint... but I wonder whether it would be possible to still show the name of the table here, with the warning sign next to it? 

I think it would be a nice hint to the user that (eg) `sample` is the table that the workflow is configured to use, but because it doesn't exist, we can't run the workflow yet. (but maybe if they imported it, we could run it?)

And I think it should still be safe to fill that value in... the user still can't submit the workflow, and it should still be obvious that something's not right.

cjllanwarne (approved):
LGTM, one last "wouldn't it be nice" level thought

src/pages/SubmissionConfig.js
kpierre13: Maybe this is a dumb thought, but wouldn't that be confusing to the user? It seems contradicting to me to tell a user, "Hey we didn't find this data table" but then show that data table in the list of imported data tables, as opposed to omitting it from the list which may then signal the user that they need to import it since it's not in the list. Maybe I could add a pop-up on the icon to suggest importing the data table?

cjllanwarne: 👍 based on our conversation this morning, I think (a) it's a lot harder and (b) I was outvoted on usefulness, so feel free to ignore this comment for now (until we get more user feedback)!

src/pages/SubmissionConfig.js
```diff
@@ -47,7 +50,7 @@ export const SubmissionConfig = ({ methodId }) => {
       const searchResult = await Ajax(signal).Wds.search.post(recordType)
       setRecords(searchResult.records)
     } catch (error) {
-      notify('error', 'Error loading WDS records', { detail: await (error instanceof Response ? error.text() : error) })
+      setNoRecordTypeData(`Data Table not found: ${recordType}`)
```
cjllanwarne, line 53: I think we're referring to a table, not the product

```suggestion
      setNoRecordTypeData(`Data table not found: ${recordType}`)
```

cjllanwarne (approved):
👍 from me on the content (except that capital T)... It occurred to me after our refinement this morning we for sure should check in with Sara that this error also makes sense.

PR #83, opened 2023-01-24
[WM-1653] Auto refresh Submission History & Details pages

salonishah11:
Tested locally:
Below screenshot (from Submission Details) shows that auto refreshing happens as long as there is a run in non-terminal state. Once all runs reach terminal state the auto refreshing stops. Same logic applies to Submission History page. 
Note: I had updated the timeout to be 1 second for testing purposes as workflows usually fail within 1 minute. console.log statements have also been removed
Screen Shot 2023-01-24 at 5 04 26 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1653

### DIFF
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -12,7 +12,7 @@ import colors from 'src/libs/colors'
 import * as Utils from 'src/libs/utils'
 
 
-export const AutoRefreshTimeout = 1000 * 60 // 1 minute
+export const AutoRefreshInterval = 1000 * 60 // 1 minute
 
 const iconSize = 24
 export const addCountSuffix = (label, count = undefined) => {
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -12,6 +12,8 @@ import colors from 'src/libs/colors'
 import * as Utils from 'src/libs/utils'
 
 
+export const AutoRefreshTimeout = 1000 * 60 // 1 minute
+
 const iconSize = 24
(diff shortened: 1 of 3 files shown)


src/components/submission-common.js
```diff
@@ -12,6 +12,8 @@ import colors from 'src/libs/colors'
 import * as Utils from 'src/libs/utils'
 
 
+export const AutoRefreshTimeout = 1000 * 60 // 1 minute
```
salonishah11, line 15: For reviewers: What should be the timeout? This value is same as we have in Terra UI. I can increase it if we think this will put too much load on CBAS.

kpierre13: My first instinct is to do whatever is done in Terra UI. Curious to see what others think as well.

kpierre13 (approved):

src/components/submission-common.js
mspector: agreed, "same as terra UI" is a good starting point until we have a reason to change it. 

it actually feels kind of long to me, from a user perspective, but that's a conversation for another time

mspector (approved):
LGTM. My only pedantic comment is that the value represented by `AutoRefreshTimeout` is more like an "interval" (and so perhaps should be called `AutoRefreshInterval`)

Mostly I mention it because there is a similar react function called `setTimeout`

salonishah11:
@mspector That makes sense. I have updated the variable name to `AutoRefreshInterval`.

PR #84, opened 2023-01-25
[WM-1661] show workflow name in tasks column when task name doesn't exist

mspector:


src/components/submission-common.js
```diff
@@ -199,8 +199,8 @@ const parseMethodString = methodString => {
   const methodNameParts = methodString.split('.')
   return {
     workflow: methodNameParts[0],
-    variable: methodNameParts[methodNameParts.length - 1],
-    call: methodNameParts.length === 3 ? methodNameParts[1] : ''
+    task: methodNameParts.length === 3 ? methodNameParts[1] : '',
```
cjllanwarne, line 202: So "call" is actually more accurate here... and would cover the concept of a workflow OR a task, so might even be better in the UI too

cjllanwarne (approved):

src/components/submission-common.js
mspector: ah, ok. I'll change that back.

PR #87, opened 2023-01-27
WM-1662: Handle UNKNOWN status for recently submitted workflows

cjllanwarne:


salonishah11 (approved):

src/pages/SubmissionDetails.js
```diff
@@ -81,7 +81,7 @@ export const SubmissionDetails = ({ submissionId }) => {
       case 'CANCELING':
         return statusType.canceling
       default:
-        return statusType.unknown
+        return (differenceFromNowInSeconds(submissionDate) < 10) ? statusType.initializing : statusType.unknown
```
salonishah11, line 84: TOL: Is 10 seconds enough? If yes, would it be worth adding a comment why 10 seconds is enough?

mspector (approved):
nice. LGTM

PR #90, opened 2023-01-30
WM-1668: Fix type names and italicize optionals

cjllanwarne:
Fixes arrays to no longer assume that the inner values must be primitives.

Stowaway changes which seemed useful to add in and took not much extra time:
- Updating optional type names from `String (optional)` to `String?`, to match how WDL displays them.
- Updating optional input font to italic on the input name and type, just like in Terra UI

Examples, including showing the table_map input now rendering correctly:
image

image


salonishah11 (approved):

mspector (approved):
👍

PR #92, opened 2023-01-31
[WM-1664] Table in Select data tab has resizable columns

salonishah11:
Data table columns are now resizable. The resized column widths are persisted within the Submission Config page for a workflow. If user navigates back to Select Workflow page and selects the previous workflow again the column widths will be reset to default. This has been purposefully kept simple for the first iteration and we can iterate on it based on user feedback.

Note: If 2 data samples have same columns, the user can resize same named columns to different widths and they will be persisted. So a column with same name across different data samples can have different column widths.

Before: Columns at default width

Resizing each columns to different width. The log messages (now removed) show how the widths are stored

Closes https://broadworkbench.atlassian.net/browse/WM-1664

### DIFF
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -67,6 +67,8 @@ export const makeStatusLine = (iconFn, label, style) => div(
 
 export const recordsTable = props => {
   const {
+    dataTableColumnWidths, setDataTableColumnWidths,
+    dataTableRef,
     records,
     selectedRecords, setSelectedRecords,
     selectedDataTable,
@@ -95,15 +97,16 @@ export const recordsTable = props => {
     return records.length && _.every(k => _.includes(k, selectedIds), recordIds)
   }
 
-  const resizeColumn = (delta, columnName) => {
-    window.alert(`column resizing currently disabled (${delta}, ${columnName}`)
+  const resizeColumn = (currentWidth, delta, columnKey) => {
+    setDataTableColumnWidths(_.set(columnKey, currentWidth + delta))
   }
 
-  const columnWidth = 300
+  const withDataTableNamePrefix = columnName => `${selectedDataTable.name}/${columnName}`
 
   return h(AutoSizer, [({ width, height }) => {
     return h(GridTable, {
       'aria-label': `${selectedDataTable.name} data table`,
+      ref: dataTableRef,
       width,
       height,
       // // Keeping these properties here as a reminder: can we use them?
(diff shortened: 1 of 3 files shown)


src/components/submission-common.js
```diff
@@ -95,15 +97,16 @@ export const recordsTable = props => {
...
+    setDataTableColumnWidths(_.set(columnKey, currentWidth + delta))
   }
 
-  const columnWidth = 300
+  const withDataTableNamePrefix = columnName => `${selectedDataTable.name}/${columnName}`
 
   return h(AutoSizer, [({ width, height }) => {
     return h(GridTable, {
       'aria-label': `${selectedDataTable.name} data table`,
+      ref: dataTableRef,
```
mspector, line 109: for my own education: what does this property do? is it easy enough to explain in a comment?

src/pages/SubmissionConfig.test.js
```diff
@@ -400,6 +400,111 @@ describe('SubmissionConfig records selector', () => {
...
+    await waitFor(() => {
+      expect(mockRunSetResponse).toHaveBeenCalledTimes(1)
+      expect(mockTypesResponse).toHaveBeenCalledTimes(1)
+      expect(mockMethodsResponse).toHaveBeenCalledTimes(0)
+      expect(mockSearchResponse).toHaveBeenCalledTimes(0)
+    })
+    const table = await screen.findByRole('table')
+    // after the initial render (not before), records data should have been retrieved once
+    await waitFor(() => {
+      expect(mockSearchResponse).toHaveBeenCalledTimes(1)
```
mspector, line 444: since nothing is happening in between these two calls to `waitFor`, I think you could put them all in a single call:

```
    await waitFor(() => {
      expect(mockRunSetResponse).toHaveBeenCalledTimes(1)
      expect(mockTypesResponse).toHaveBeenCalledTimes(1)
      expect(mockMethodsResponse).toHaveBeenCalledTimes(1)
      expect(mockSearchResponse).toHaveBeenCalledTimes(1)
    })
```

mspector (approved):
I checked it out/ tried it and it works great! thanks for adding the screenshot

I left a question for you, and a non-critical suggestion in the test

kpierre13 (approved):

src/components/submission-common.js
salonishah11: So from what I understand, `useRef()` is a hook and it returns a mutable object. The initial value for the object is stored in `current` attribute. We can pass the reference to this object using `ref` variable and what this means is that whenever the corresponding DOM node's value changes, react will update the `.current` attribute for the ref. 
Reference: https://reactjs.org/docs/hooks-reference.html#useref

PR #94, opened 2023-01-31
[WM-1667] Don't submit parameter_value when input source is changed to None

salonishah11:
This PR fixes the behavior mentioned in the ticket. 
Before: The payload being sent for input switched to "None" source included `parameter_value`
```
{
  "input_name": "fetch_sra_to_bam.Fetch_SRA_to_BAM.docker",
  "input_type": {
    "type": "optional",
    "optional_type": {
      "type": "primitive",
      "primitive_type": "String"
    }
  },
  "source": {
    "type": "none",
    "parameter_value": "quay.io/broadinstitute/ncbi-tools:2.10.7.10"
  }
}
```

Now
```
{
    "input_name": "fetch_sra_to_bam.Fetch_SRA_to_BAM.docker",
    "input_type": {
      "type": "optional",
      "optional_type": {
        "type": "primitive",
        "primitive_type": "String"
      }
    },
    "source": {
      "type": "none"
    }
}
```

Closes https://broadworkbench.atlassian.net/browse/WM-1667

### DIFF
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -330,7 +330,7 @@ export const inputsTable = props => {
                   const param = newType === 'record_lookup' ? 'record_attribute' : 'parameter_value'
                   newSource = {
                     type: newType,
-                    [param]: _.get(`${rowIndex}.source.${param}`, configuredInputDefinition)
+                    [param]: ''
                   }
                 }
                 const newConfig = _.set(`${rowIndex}.source`, newSource, configuredInputDefinition)
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -321,10 +321,17 @@ export const inputsTable = props => {
               value: _.get(_.get(`${rowIndex}.source.type`, configuredInputDefinition), inputSourceLabels) || null,
               onChange: ({ value }) => {
                 const newType = _.get(value, inputSourceTypes)
-                const param = newType === 'record_lookup' ? 'record_attribute' : 'parameter_value'
-                const newSource = {
-                  type: newType,
-                  [param]: _.get(`${rowIndex}.source.${param}`, configuredInputDefinition)
+                let newSource
+                if (newType === 'none') {
+                  newSource = {
+                    type: newType
+                  }
+                } else {
+                  const param = newType === 'record_lookup' ? 'record_attribute' : 'parameter_value'
+                  newSource = {
+                    type: newType,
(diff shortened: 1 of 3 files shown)


src/pages/SubmissionConfig.js
```diff
@@ -42,7 +42,7 @@ export const SubmissionConfig = ({ methodId }) => {
   const [inputTableSort, setInputTableSort] = useState({ field: 'taskVariable', direction: 'asc' })
   const [outputTableSort, setOutputTableSort] = useState({ field: 'taskVariable', direction: 'asc' })
 
-  const [launching, setLaunching] = useState(undefined)
+  const [displayLaunchModal, setDisplayLaunchModal] = useState(false)
```
mspector, line 45: 👍

mspector (approved):
Looks good to me.

I had a thought while reviewing this: could this be unit tested instead of requiring a relatively complex UI test?

Answering my own question: it seems like probably a UI test *is* necessary here, because this `onChange` function uses global state... but I wonder if it's worth thinking about different ways to approach this, so that it is unit-testable.

salonishah11:
@mspector Yes I though about some other way to write this test but I wasn't able to since the behavior I wanted to test depends on state changes which I wasn't sure how to do in a smaller unit testable way.

src/components/submission-common.js
```diff
@@ -321,10 +321,17 @@ export const inputsTable = props => {
...
+                let newSource
+                if (newType === 'none') {
+                  newSource = {
+                    type: newType
+                  }
+                } else {
+                  const param = newType === 'record_lookup' ? 'record_attribute' : 'parameter_value'
+                  newSource = {
+                    type: newType,
+                    [param]: _.get(`${rowIndex}.source.${param}`, configuredInputDefinition)
```
cjllanwarne, line 333: Just thinking this through... If we're changing the param type, shouldn't we unset this? This looks like it would either be empty anyway (because param is changing), or store multiple values in the `configuredInputDefinition` so that it can be switched back to... but neither of those sound better than just setting this to `''` here and keeping it simple?

cjllanwarne (approved):
Looks functionally correct. I just have one question about what state is beings set to, and whether all the state-storage is actually doing anything

src/components/submission-common.js
salonishah11: Hmm good point. Yeah it is basically empty in either case so it makes sense to simply set it to `''` here. I will update it 👍

PR #96, opened 2023-02-07
[WM-1698] configurable instance id

mspector:


public/config.json
```diff
@@ -1,5 +1,6 @@
 {
   "cbasUrlRoot" : "http://localhost:3000",
   "cromwellUrlRoot" : "http://localhost:3000/cromwell",
-  "wdsUrlRoot" : "http://localhost:3000/wds"
+  "wdsUrlRoot" : "http://localhost:3000/wds",
+  "wdsInstanceId": "15f36863-30a5-4cab-91f7-52be439f1175"
```
jgainerdewar, line 5: Can we set an empty string here rather than this UUID?

src/libs/ajax.js
```diff
@@ -109,8 +109,8 @@ const Cromwell = signal => ({
 })
 
 // TODO: how to get these values?
-const wdsInstanceId = '15f36863-30a5-4cab-91f7-52be439f1175'
-const wdsApiVersion = 'v0.2'
+const wdsInstanceId = getConfig().wdsInstanceId
```
cjllanwarne, line 112: Could we do the same as the api version, like `getConfig().wdsInstanceId ||  '15f36863-30a5-4cab-91f7-52be439f1175'`? To keep it all working nicely locally without setting anything up in the config?

cjllanwarne (approved):

src/libs/ajax.js
mspector: I don't mind, but it occurs to me that the randomness of this UUID could be misleading. What if we used `00000000-0000-0000-0000-000000000000` instead? 

or does changing the UUID defeat the purpose?

mspector: I'm just going to add it, for the sake of speed... it's such a minor issue

jgainerdewar (approved):

PR #100, opened 2023-02-15
WM-1733: Run CI on addition to merge group

cjllanwarne:


.github/workflows/publish.yml
```diff
@@ -22,6 +22,10 @@ on:
     branches: [ '**' ]
     paths-ignore:
       - 'README.md'
+  merge_group:
```
salonishah11, line 25: do we want this here? 🤔

salonishah11: If this is needed here as well then we should add `merge_group` to [publish.yaml](https://github.com/DataBiosphere/cbas/blob/main/.github/workflows/publish.yml) in CBAS repo as well

salonishah11 (approved):

kpierre13 (approved):

.github/workflows/publish.yml
cjllanwarne: As discussed in person, just following wherever I see actions being triggered by `on: pull_request`

PR #101, opened 2023-02-15
[WM-1717] Display correct duration for Run Set in Submission Details page

salonishah11:
Tested locally:

Closes https://broadworkbench.atlassian.net/browse/WM-1717

### DIFF
--- a/src/pages/SubmissionDetails.js
+++ b/src/pages/SubmissionDetails.js
@@ -6,14 +6,14 @@ import { AutoSizer } from 'react-virtualized'
 import { ButtonPrimary, Link, Navbar, Select } from 'src/components/common'
 import { HeaderSection, statusType, SubmitNewWorkflowButton } from 'src/components/job-common'
 import Modal from 'src/components/Modal'
-import { AutoRefreshInterval, makeStatusLine } from 'src/components/submission-common'
+import { AutoRefreshInterval, getDuration, isRunInTerminalState, isRunSetInTerminalState, makeStatusLine } from 'src/components/submission-common'
 import { FlexTable, paginator, Sortable, tableHeight, TextCell } from 'src/components/table'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import { goToPath } from 'src/libs/nav'
 import { notify } from 'src/libs/notifications'
 import { useCancellation, useOnMount } from 'src/libs/react-utils'
-import { customFormatDuration, differenceFromDatesInSeconds, differenceFromNowInSeconds, makeCompleteDate } from 'src/libs/utils'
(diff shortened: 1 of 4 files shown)


src/pages/SubmissionDetails.js
```diff
@@ -179,7 +166,7 @@ export const SubmissionDetails = ({ submissionId }) => {
         h2(['Submission name: ', specifyRunSet[0]?.run_set_name]),
         h3(['Workflow name: ', getSpecificMethod[0]?.name]),
         h3(['Submission date: ', specifyRunSet[0] && makeCompleteDate(specifyRunSet[0].submission_timestamp)]),
-        h3(['Duration: ', specifyRunSet[0] && customFormatDuration(duration(specifyRunSet[0]))])
+        h3(['Duration: ', specifyRunSet[0] && customFormatDuration(getDuration(specifyRunSet[0].state, specifyRunSet[0].submission_timestamp, specifyRunSet[0].last_modified_timestamp, isRunSetInTerminalState))])
```
salonishah11, line 169: This is the fix for the bug listed in the ticket. We were incorrectly using `isRunInTerminalState` to check if Run Set was in terminal state. 

Other changes are to reduce code redundancy around getting duration value and making it more easily testable.

kpierre13 (approved):

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -182,7 +169,8 @@ export const SubmissionHistory = () => {
                   field: 'duration',
                   headerRenderer: () => h(Sortable, { sort, field: 'duration', onSort: setSort }, ['Duration']),
                   cellRenderer: ({ rowIndex }) => {
-                    return h(TextCell, [Utils.customFormatDuration(runSetDuration(paginatedPreviousRunSets[rowIndex]))])
+                    const row = paginatedPreviousRunSets[rowIndex]
+                    return h(TextCell, [Utils.customFormatDuration(getDuration(row.state, row.submission_timestamp, row.last_modified_timestamp, isRunSetInTerminalState))])
```
cjllanwarne, line 173: How come this one needs a custom duration format vs the others locations?

cjllanwarne (approved):

src/pages/SubmissionHistory/SubmissionHistory.js
salonishah11: So whenever we want to display duration we use custom format to make a human readable format as compared to [this line](https://github.com/DataBiosphere/cbas-ui/blob/86f4d6d3d5d939e11a85a8f092b27a62e02f2de9/src/pages/SubmissionHistory/SubmissionHistory.js#L30) where we just calculate duration per run set and save it back to run set data.

PR #102, opened 2023-02-16
[WM-1715] loading data spinners

mspector:


src/pages/SubmissionDetails.test.js
```diff
@@ -102,19 +102,6 @@ describe('Submission Details page', () => {
     Object.defineProperty(HTMLElement.prototype, 'offsetWidth', { configurable: true, value: 800 })
   })
 
-  beforeEach(() => {
-    const getRunsMethod = jest.fn(() => Promise.resolve(runsData))
```
mspector, line 106: this change has implications throughout the test, but is necessary for testing the case where `runsData` should be empty.

salonishah11: What happens if you keep this here, but override it's implementation locally inside the test when you want to return empty `runsData`?

src/pages/SubmissionDetails.js
```diff
@@ -128,7 +128,9 @@ export const SubmissionDetails = ({ submissionId }) => {
       }
     }
 
-    loadRunSetData().then(runSet => runSet && loadMethodsData(runSet.method_version_id))
+    setLoading(true)
```
salonishah11, line 131: I think it would be better to use [withBusyState](https://github.com/DataBiosphere/cbas-ui/blob/8798c52740ddca0dbf0715661bb0ecdbc24cea1f/src/libs/utils.js#L135-L142) instead of manually setting the flag. Once it is set to true only refreshing the component will set it back to false. `withBusyState` will do that for you - set the flag to true, wait for the passed function to finish and then set the flag back to false. 

Link to a usage example in [JobHistory page in Terra UI](https://github.com/DataBiosphere/terra-ui/blob/f7964e84f2d1ef3d79ac5ad842cf2deb5685043d/src/pages/workspaces/workspace/JobHistory.js#L137).

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -53,7 +54,8 @@ export const SubmissionHistory = () => {
   }
 
   useOnMount(async () => {
-    await refresh()
+    setLoading(true)
+    await refresh().then(() => setLoading(false))
```
salonishah11, line 58: Same comment about `withBusyState` here as well

src/pages/SubmitWorkflow.js
```diff
@@ -42,12 +43,12 @@ export const SubmitWorkflow = () => {
         notify('error', 'Error loading saved workflows', { detail: await (error instanceof Response ? error.text() : error) })
       }
     }
-
+    setLoading(true)
```
salonishah11, line 46: Same comment about `withBusyState` here as well

cjllanwarne (approved):
Looks good for submission details. Do we also want to address loading on the Submission History, Submission Config and Landing Page?

src/pages/SubmissionDetails.test.js
mspector: you know what, that seems to work just fine. good catch.

given that, do you think I should restore the `beforeEach`? I don't feel strongly about it, but I kind of like having each test set itself up explicitly (I'm less particular about DRY when it comes to tests, personally)

src/pages/SubmissionDetails.js
mspector: this looks cool - I can try this

mspector:
> Looks good for submission details. Do we also want to address loading on the Submission History, Submission Config and Landing Page?

Submission History and the landing page (Submit Workflow) have the new loading spinners, there just aren't as many changes to their tests because of the specifics around how each were originally written. I intentionally didn't add tests to check for loading spinners because I felt that they're more like decoration than actual functionality, so I just made sure that the existing tests took the spinners into account when constructing the page.

I don't think Submission Config needed these, but I will double check when I apply Saloni's suggestions

src/pages/SubmissionDetails.test.js
salonishah11: I think it would make sense to restore `beforeEach` because we use the same setup for almost all tests (and explicitly change it inside one test). That being said I don't have any strong opinion either way so I would be happy whichever way you decide to implement it.

salonishah11 (approved):

PR #116, opened 2023-03-01
[WM-1743] Initial layout for "Find Workflow" Modal

salonishah11:
Note to Reviewers: Both `ModalDrawer.js` and `MethodCard.js` are copy pasted from Terra UI and tweaked for styling and displaying content.

Screenshot:
Clicking on Find Workflow button opens the modal drawer with "Browse Suggested workflows" selected and 3 hard coded method cards

Screen Shot 2023-03-01 at 4 16 39 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1743


src/libs/style.js
```diff
@@ -88,23 +88,23 @@ export const cardList = {
...
-export const navList = {
+export const findWorkflowNavList = {
   heading: {
     color: colors.dark(), backgroundColor: colors.light(0.4), fontSize: 16, padding: '1rem 1.5rem',
     display: 'flex', justifyContent: 'space-between', alignItems: 'center', margin: 0,
     fontWeight: 600, textTransform: 'uppercase', borderBottom: `0.5px solid ${colors.dark(0.2)}`
   },
   itemContainer: selected => ({
     display: 'flex', alignItems: 'center', flex: 'none', width: '100%', height: 50,
-    padding: '0 1.5rem', backgroundColor: 'white',
```
cjllanwarne, line 99: Will this have unintended consequences in our other workflow nav lists? (presumably... just the main landing page?)

cjllanwarne (approved):
💯 !

src/libs/style.js
salonishah11: It turned out that this `navList` styling wasn't used in CBAS UI anywhere (yet). Hence I just removed these lines. My assumption is in future when we do need it we can have general `navList` style again with `findWorkflowNavList` only editing what it doesn't need.

PR #118, opened 2023-03-02
WX-956 Add workflow engine is to run details page

jgainerdewar:
Screen Shot 2023-03-02 at 4 12 02 PM

### DIFF
--- a/src/pages/RunDetails.test.js
+++ b/src/pages/RunDetails.test.js
@@ -20,7 +20,7 @@ const runDetailsProps = {
   namespace: 'example-billing-project',
   name: 'workspace',
   submissionId: 1,
-  workflowId: '5d96fd3c-1a89-40ae-8095-c364181cda46'
+  workflowId: '00001111-2222-3333-aaaa-bbbbccccdddd'
 }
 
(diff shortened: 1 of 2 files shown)


src/pages/RunDetails.test.js
```diff
@@ -20,7 +20,7 @@ const runDetailsProps = {
   namespace: 'example-billing-project',
   name: 'workspace',
   submissionId: 1,
-  workflowId: 2
+  workflowId: '5d96fd3c-1a89-40ae-8095-c364181cda46'
```
aednichols, line 23: I once lost several hours trying to find the origin of a real-looking but hardcoded workflow ID and now believe in fake-looking IDs
```suggestion
  workflowId: '00001111-2222-3333-aaaa-bbbbccccdddd'
```

aednichols (approved):

src/pages/RunDetails.test.js
jgainerdewar: Totally 👍 This needs to be updated in two places, so will push a separate commit.

THWiseman (approved):

PR #119, opened 2023-03-02
[WM-1755] Fix auto-refresh on Submission Details page

kpierre13:
Screenshots show the duration column when first submitted, after a minute, and after 2 minutes. Both the run durations and run set durations give the correct time when being auto-refreshed.

image
image
image

### DIFF
--- a/src/pages/SubmissionDetails.js
+++ b/src/pages/SubmissionDetails.js
@@ -6,7 +6,7 @@ import { ButtonPrimary, Link, Navbar, Select } from 'src/components/common'
 import { centeredSpinner } from 'src/components/icons'
 import { HeaderSection, statusType, SubmitNewWorkflowButton } from 'src/components/job-common'
 import Modal from 'src/components/Modal'
-import { AutoRefreshInterval, getDuration, isRunInTerminalState, isRunSetInTerminalState, makeStatusLine, loadRunSetData, loadRunsData } from 'src/components/submission-common'
+import { AutoRefreshInterval, getDuration, isRunInTerminalState, isRunSetInTerminalState, makeStatusLine, loadRunSetData } from 'src/components/submission-common'
 import { FlexTable, paginator, Sortable, tableHeight, TextCell } from 'src/components/table'
 import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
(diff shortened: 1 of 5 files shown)


src/pages/SubmissionDetails.test.js
```diff
@@ -151,7 +151,7 @@ describe('Submission Details page', () => {
 
     await waitFor(() => {
       expect(getRuns).toHaveBeenCalledTimes(1)
-      expect(getRunsSets).toHaveBeenCalledTimes(1)
+      expect(getRunsSets).toHaveBeenCalledTimes(3)
```
salonishah11, line 154: Out of curiosity, why is this being called 3 times now? 🤔

kpierre13: In my most recent commit, this will be called twice now because it's called once in the `refresh()` function, and then called again when mounted. Is this too many? Should I try and consolidate?

salonishah11: No I was just curious. Sounds good 👍

src/pages/SubmissionDetails.js
```diff
@@ -81,28 +81,22 @@ export const SubmissionDetails = ({ submissionId }) => {
...
+      const loadedRunSetData = await loadRunSetData(signal)
+      setRunSetData(loadedRunSetData)
+
+      // only refresh if there are Run Sets/Runs in non-terminal state
+      if (_.some(({ state }) => !isRunSetInTerminalState(state), loadedRunSetData)) {
+        // Nesting these checks to avoid a "double refresh"
+        if (_.some(({ state }) => !isRunInTerminalState(state), runsAnnotatedWithDurations)) {
+          scheduledRefresh.current = setTimeout(refresh, AutoRefreshInterval)
+        }
       }
```
salonishah11, line 93: This can be consolidated to checking for non-terminal state for Run Set only. This way we refresh until Run Set reaches terminal state. Otherwise in this conditional logic it could be possible that all Runs are in terminal state but the Run Set isn't and as a result we don't refresh the page at all.

cjllanwarne (approved):
LGTM with Saloni's suggestion

src/pages/SubmissionDetails.js
```diff
@@ -81,8 +81,11 @@ export const SubmissionDetails = ({ submissionId }) => {
       const runsAnnotatedWithDurations = _.map(r => _.merge(r, { duration: getDuration(r.state, r.submission_date, r.last_modified_timestamp, isRunInTerminalState) }), runs)
       setRunsData(runsAnnotatedWithDurations)
 
-      // only refresh if there are Runs in non-terminal state
-      if (_.some(({ state }) => !isRunInTerminalState(state), runs)) {
+      const loadedRunSetData = await loadRunSetData(signal)
+      setRunSetData(loadedRunSetData)
+
+      // only refresh if there are Run Sets/Runs in non-terminal state
```
salonishah11, line 87: small nit: this can now be updated to `only refresh if the Run Set is in non-terminal state`

salonishah11 (approved):

PR #120, opened 2023-03-03
WX-957 Remove call cache result column

jgainerdewar:
Screen Shot 2023-03-03 at 1 28 11 PM

I removed the call caching related bits from the components the user sees, but left the now-unused CallCachingWizard in place since we'll be enabling call caching sometime in the near future. 

Note, the Links column is empty in this screenshot but contains stuff if the task fails.

### DIFF
--- a/src/pages/workspaces/workspace/jobHistory/CallTable.js
+++ b/src/pages/workspaces/workspace/jobHistory/CallTable.js
@@ -1,5 +1,5 @@
 import _ from 'lodash/fp'
-import { Fragment, useState } from 'react'
+import { useState } from 'react'
 import { div, h } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
 import { Link, Select } from 'src/components/common'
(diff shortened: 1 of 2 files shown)


src/pages/workspaces/workspace/jobHistory/CallTable.js
```diff
@@ -137,8 +109,7 @@ const CallTable = ({ namespace, name, submissionId, workflowId, callName, callOb
         ]
       })
     ]),
-    failuresModalParams && h(FailuresModal, { ...failuresModalParams, callFqn: callName, onDismiss: () => setFailuresModalParams(undefined) }),
-    wizardSelection && h(CallCacheWizard, { onDismiss: () => setWizardSelection(undefined), namespace, name, submissionId, workflowId, ...wizardSelection })
```
cjllanwarne, line 141: Presumably there is a `CallCacheWizard` element somewhere that would also benefit from removal?

cjllanwarne (approved):

kraefrei (approved):
LGTM

PR #122, opened 2023-03-06
[WM-1744] First WARP workflow in the UI

kpierre13:
Method loaded into the config page after selecting:
image

Method present in workflow cards list after importing:
image

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -92,6 +92,10 @@ const Cbas = signal => ({
       const keyParams = qs.stringify({ method_version_id: methodVersionId })
       const res = await fetchCbas(`methods?${keyParams}`, { signal, method: 'GET' })
       return await res.json()
+    },
+    post: async payload => {
+      const res = await fetchCbas(`methods`, _.mergeAll([{ signal, method: 'POST' }, jsonBody(payload)]))
+      return res.json()
     }
   }
 })
--- a/src/pages/FindWorkflow/FindWorkflowModal.js
+++ b/src/pages/FindWorkflow/FindWorkflowModal.js
@@ -29,7 +29,8 @@ const suggestedWorkflowsList = [
   {
     method_name: 'Optimus',
     method_description: 'The optimus 3 pipeline processes 10x genomics sequencing data based on the v2 chemistry. It corrects cell barcodes and UMIs, aligns reads, marks duplicates, and returns data as alignments in BAM format and as counts in sparse matrix exchange format.',
-    method_source: 'Dockstore',
+    // TODO: Change to Dockstore when we accept Dockstore imports
+    method_source: 'GitHub',
     method_version: 'Optimus_v5.5.0',
     method_url: 'https://raw.githubusercontent.com/broadinstitute/warp/develop/pipelines/skylab/optimus/Optimus.wdl'
   },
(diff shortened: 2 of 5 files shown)


src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -22,11 +27,11 @@ const styles = {
 
 const suggestedWorkflowsList = [
   {
-    method_name: 'mock_method_1',
-    method_description: 'mock_method_1 description. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.',
+    method_name: 'Optimus',
+    method_description: 'The optimus 3 pipeline processes 10x genomics sequencing data based on the v2 chemistry. It corrects cell barcodes and UMIs, aligns reads, marks duplicates, and returns data as alignments in BAM format and as counts in sparse matrix exchange format.',
     method_source: 'GitHub',
```
salonishah11, line 32: I think the source for this is [Dockstore](https://dockstore.org/workflows/github.com/broadinstitute/warp/Optimus:Optimus_v5.5.0?tab=info) based on the workflow config page in Terra

salonishah11 (approved):

src/pages/FindWorkflow/FindWorkflowModal.js
kpierre13: This means that we won't be able to import this method right? Is there a check in CBAS that stops non-Github workflows from being imported?

cjllanwarne:
We probably want to update that `"Data table not found: undefined"` warning too, to just say `"Select a data table"`?

src/pages/FindWorkflow/FindWorkflowModal.js
salonishah11: Ah you are right. We won't be able to import if the source is Dockstore. I guess then we can leave it as Github for now and change it when we add Dockstore integration.

src/libs/ajax.js
```diff
@@ -92,6 +92,10 @@ const Cbas = signal => ({
       const keyParams = qs.stringify({ method_version_id: methodVersionId })
       const res = await fetchCbas(`methods?${keyParams}`, { signal, method: 'GET' })
       return await res.json()
+    },
+    post: async payload => {
```
cjllanwarne, line 96: TOL: For future usage and maintainability, this might be a little easier to work with if we spell out the parameters and construct the payload inside the method.

src/pages/FindWorkflow/FindWorkflowModal.js
cjllanwarne: Right now this "source" is informational only, so we can put whatever value we want there

cjllanwarne: We can check with Sara when she's back

src/pages/SubmissionConfig.js
```diff
@@ -58,7 +58,7 @@ export const SubmissionConfig = ({ methodId }) => {
       const searchResult = await Ajax(signal).Wds.search.post(recordType)
       setRecords(searchResult.records)
     } catch (error) {
-      setNoRecordTypeData(`Data table not found: ${recordType}`)
+      setNoRecordTypeData('Select a data table')
```
cjllanwarne, line 61: So I think we want to keep the `Data table not found: ${recordType}` if there is a `recordType` to show. But if it's null or undefined, we probably want to show the `'Select a data table'` version.

cjllanwarne (approved):

PR #123, opened 2023-03-07
[WM-1745] Nonstruct workflows in UI

kpierre13:


src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -34,18 +34,32 @@ const suggestedWorkflowsList = [
...
+    method_source: 'GitHub',
+    method_version: 'MultiSampleSmartSeq2SingleNuclei_v1.2.14',
+    method_url: 'https://raw.githubusercontent.com/broadinstitute/warp/develop/pipelines/skylab/smartseq2_single_nucleus_multisample/MultiSampleSmartSeq2SingleNucleus.wdl'
+  },
+  {
+    method_name: 'scATAC',
+    method_description: 'Processing of single-cell ATAC-seq data with the scATAC pipeline.',
+    method_source: 'GitHub',
+    method_version: 'scATAC 1.2.0',
+    method_url: 'https://raw.githubusercontent.com/broadinstitute/warp/develop/pipelines/skylab/scATAC/scATAC.wdl'
```
salonishah11, line 62: For my own clarification - this list also cover workflows with struct data type?

kpierre13: Only workflows without structs

salonishah11: I see. I thought since 4 new workflows are being added few of them must have Struct data types since we are only going to support 5 WARP workflows in the first iteration. Maybe I misunderstood something.

salonishah11 (approved):

src/pages/FindWorkflow/FindWorkflowModal.js
kpierre13: Well, there are 7 workflows in the excel doc. 2 of them say not for the first iteration, but I thought that was because they're trivial? If we don't want them shown to users at all I can remove them entirely.

salonishah11: Ah I see. My understanding was, and it could be wrong, is that we won't be adding support for those 2 workflows in the first iteration. But it would be worth checking with Sara or @cjllanwarne.

PR #128, opened 2023-03-13
[WM-1763] Get WDS url root from Leo

salonishah11:
Note: CBAS UI will talk to Leo only when `wdsUrlRoot` is not present in config. When we are ready for CBAS UI to talk to Leo to get WDS url root, we should remove`wdsUrlRoot` from config and CBAS UI will start pinging Leo to get the url root.

Locally:
CBAS UI doesn't talk to Leo and directly talks to WDS to load data

-----

In App:
I manually tested this by removing `wdsUrlRoot` and Leo dev url from the cromwhelm config and deploying the new helm chart to app running in Dev. Since `wdsUrlRoot` is not present in config, CBAS UI pings Leo to get WDS url and then loads the data using that url

Closes https://broadworkbench.atlassian.net/browse/WM-1763


src/pages/SubmissionConfig.js
```diff
@@ -97,19 +130,45 @@ export const SubmissionConfig = ({ methodId }) => {
...
+    }
+  }, [loadRecordsData, loadRecordTypes, loadWdsUrl, wdsProxyUrl])
 
   useOnMount(() => {
-    loadTablesData()
     loadRunSet().then(runSet => {
+      setRunSetRecordType(runSet.record_type)
       loadMethodsData(runSet.method_id, runSet.method_version_id)
-      loadRecordsData(runSet.record_type)
+      loadWdsData(runSet.record_type, true)
```
mspector, line 171: I'm not sure what the best practice is here, but I find this second boolean argument a little unintuitive. Does it make sense to have this function accept an object as its argument, so that there can be some indication of what the function is doing at the place in the code where it's being called?

I'm imagining something like: 
`loadWdsData({ recordType: runSet.record_type, includeLoadRecordTypes: true })` 

(this might be a bad idea for some reason that a JS expert would identify...)

mspector (approved):
a few questions and one small readability suggestion that you can take or leave as you see fit. LGTM

src/pages/SubmissionConfig.js
```diff
@@ -97,19 +130,45 @@ export const SubmissionConfig = ({ methodId }) => {
...
+  const loadWdsData = useCallback(async (recordType, isLoadRecordTypes) => {
+    try {
+      // try to load WDS proxy URL if one doesn't exist
+      if (!wdsProxyUrl || (wdsProxyUrl.status !== 'Ready')) {
+        const { status, state: wdsUrlRoot } = await loadWdsUrl()
+        if (status === 'Unauthorized') {
+          notify('warn', 'Error loading data tables', { detail: 'Service returned Unauthorized error. Session might have expired. Please close the tab and re-open it.' })
+        } else if (!!wdsUrlRoot) {
+          if (isLoadRecordTypes) await loadRecordTypes(wdsUrlRoot)
+          await loadRecordsData(recordType, wdsUrlRoot)
```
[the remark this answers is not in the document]
salonishah11: So in single line if statement, only the first line after the if condition gets executed. So in this case if `isLoadRecordTypes` is true, only `await loadRecordTypes(wdsUrlRoot)` on L149 will be executed. I can add the curly braces is it helps make this more readable?
```
....
if (isLoadRecordTypes) { await loadRecordTypes(wdsUrlRoot) }
await loadRecordsData(recordType, wdsUrlRoot)
....
```

src/pages/SubmissionConfig.js
salonishah11: I don't think that's a bad practice. I have seen usage of passing objects/named parameters to functions as compared to positional arguments in Terra UI (as well as CBAS UI?). Since it improves readability of the code I will update the function to use named parameters 👍

kpierre13 (approved):

PR #129, opened 2023-03-14
[WM-1741] struct builder modal

mspector:


src/libs/utils.js
```diff
@@ -182,5 +182,8 @@ export const renderTypeText = iotype => {
   if (_.get('type', iotype) === 'map') {
     return `Map[${_.get('key_type', iotype)}, ${renderTypeText(_.get('value_type', iotype))}]`
   }
+  if (_.get('type', iotype) === 'struct') {
+    return `Struct` // TODO
```
mspector, line 186: is this how struct types should be rendered in the "type" column?

PR #130, opened 2023-03-14
[WM-1741] [WM-1742] struct builder modal

mspector:


src/components/InputsTable.js
```diff
@@ -0,0 +1,205 @@
...
+      onClick: () => {
+        setStructBuilderVisible(true)
+        setStructBuilderPath([rowIndex])
+      }
+    })
+  }
+
+  return h(AutoSizer, [({ width, height }) => {
+    return h(div, {}, [
+      structBuilderVisible ? h(StructBuilderModal, {
```
salonishah11, line 125: This can be updated to `structBuilderVisible && h()` so that you don't have to do the ternary condition.

src/components/InputsTable.js
```diff
@@ -0,0 +1,205 @@
...
+  StructBuilderLink
+} from 'src/components/submission-common'
+import { FlexTable, HeaderCell, Sortable, TextCell } from 'src/components/table'
+import TooltipTrigger from 'src/components/TooltipTrigger'
+import colors from 'src/libs/colors'
+import * as Utils from 'src/libs/utils'
+import { StructBuilderModal } from 'src/pages/StructBuilderModal'
+
+
+const buildStructPath = ({ structBuilderPath, headTemplate, pathTemplate, lastTemplate }) => {
```
salonishah11, line 20: I think to help better understand, it might be worth adding comments about what these 4 functions do, maybe with examples?

src/components/submission-common.js
```diff
@@ -94,6 +94,48 @@ export const loadRunSetData = async signal => {
   }
 }
 
+// Invokes logic to determine the appropriate app for WDS
+// If WDS is not running, a URL will not be present, in this case we return empty string
+// Note: This logic has been copied from how DataTable finds WDS app in Terra UI (https://github.com/DataBiosphere/terra-ui/blob/ac13bdf3954788ca7c8fd27b8fd4cfc755f150ff/src/libs/ajax/data-table-providers/WdsDataTableProvider.ts#L94-L147)
+export const resolveWdsApp = apps => {
```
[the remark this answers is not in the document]
mspector: seems to be no longer an issue

src/components/InputsTable.js
```diff
@@ -140,15 +93,9 @@ const InputsTable = props => {
   return h(AutoSizer, [({ width, height }) => {
     return h(div, {}, [
       structBuilderVisible && h(StructBuilderModal, {
-        structBuilderName: _.get(buildStructNamePath(structBuilderPath), inputTableData),
-        structBuilderBreadcrumbs: buildStructBuilderBreadcrumbs(structBuilderPath, inputTableData),
-        structBuilderInputType: _.get(buildStructInputTypePath(structBuilderPath), inputTableData),
+        inputTableData,
```
cjllanwarne, line 96: Nice! This feels a lot neater to me!

Happy to approve this now, with one final (non-critical, TOL-level) thought: `inputTableData` feels very "grab bag of stuff"-y to me. Are you using all of it, or it is only a specific subset you really need? (eg, could you replace this passing of the entire `inputTableData` object with only the fields that the struct builder really needs (maybe just the set of struct paths for this particular struct?)

cjllanwarne (approved):

src/components/InputsTable.js
mspector: The funny thing is: this exact thought is what caused me to put all of those `buildStructPath` functions in the InputTable component in the first place. The purpose of those functions is to pull out the parts of `inputTableData` that the struct builder needs.

But actually, I think there may be an in-between that will achieve what you're suggesting *and* make the `buildStructPath` functions slightly less complicated. I'll give it a try in the time between now and sprint planning...

salonishah11 (approved):

PR #131, opened 2023-03-14
WM-1795: Alert on incomplete status polls

cjllanwarne:


src/pages/SubmissionDetails.js
```diff
@@ -104,8 +106,7 @@ export const SubmissionDetails = ({ submissionId }) => {
       }
     }
 
-    await refresh()
-    loadRunSetData(signal).then(runSet => runSet && loadMethodsData(runSet.method_version_id))
+    await refresh().then(() => loadMethodsData(undefined))
```
cjllanwarne, line 109: This might look like a very dramatic change, but if you look at the actual calls in the browser, you'll notice that `runSet.method_version_id` was always undefined... presumably because `loadRunSetData` is actually returning an array, and the array doesn't have a `method_version_id` field

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -26,8 +28,9 @@ export const SubmissionHistory = () => {
   // helper for auto-refresh
   const refresh = Utils.withBusyState(setLoading, async () => {
     try {
-      const loadedRunSetData = await loadRunSetData(signal)
+      const loadedRunSetData = await loadAllRunSets(signal)
       setRunSetData(loadedRunSetData)
+      setRunSetsFullyUpdated(true)
```
salonishah11, line 33: Should this be set based on the `fully_updated` value returned in run set similar to one in Submission Details page?

cjllanwarne: Oops, it definitely should, good spot!

src/pages/SubmissionDetails.test.js
```diff
@@ -460,4 +461,63 @@ describe('Submission Details page', () => {
...
+
+    // Act
+    await act(async () => {
+      await render(h(SubmissionDetails))
+    })
+
+    expect(screen.getByText('Workflow statuses are all up to date.')).toBeInTheDocument()
+  })
+
+  it('should indicate fully updated polls', async () => {
```
salonishah11, line 504: nit: Can this be changed to `should indicate runs are not fully updated`?

salonishah11 (approved):

PR #134, opened 2023-03-19
[WM-1796] Modal dismissed on error

kpierre13:


src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -87,6 +87,7 @@ const FindWorkflowModal = ({ onDismiss }) => {
       })
     } catch (error) {
       notify('error', 'Error creating new method', { detail: await (error instanceof Response ? error.text() : error) })
+      onDismiss()
```
cjllanwarne, line 90: Thinking out loud:

This is an interesting dilemma. By dismissing the modal when something goes wrong, the user has no chance to "change something and try again", instead they presumably have to start everything all over again from scratch.

I guess one other option would be to set a piece of error message state when something goes wrong, and integrate the error message into the modal itself?

In this case it's not super bad because the only thing a user had to do was to click a button, so this is certainly not a blocker now... but once we get into more complex import modes and options, sending users back to the very start might be a little annoying.

salonishah11: That's a good point. Maybe it is worth looking into how we can display the notification banner on top of the Find Workflow modal? This way we won't need to dismiss the modal and users will be able to update things more easily

kpierre13: I created a ticket [here](https://broadworkbench.atlassian.net/browse/WM-1831) for this work since it isn't incredibly urgent. I'm also planning on following up in the terra-ui channel to see if this exists in Terra and is possible.

salonishah11 (approved):

PR #137, opened 2023-03-27
[WM-1837] Provide Covid-19 tutorial links for user in Submission Config page

salonishah11:
For 3 Covid-19 workflows the Have Questions box will display below 3 links:
Screenshot 2023-03-27 at 5 23 05 PM

For all other workflows only this link will be displayed:
Screenshot 2023-03-27 at 5 23 23 PM

Closes https://broadworkbench.atlassian.net/browse/WM-1837

### DIFF
--- a/src/pages/SubmissionConfig.js
+++ b/src/pages/SubmissionConfig.js
@@ -1,7 +1,8 @@
 import _ from 'lodash/fp'
 import { Fragment, useCallback, useEffect, useRef, useState } from 'react'
-import { a, div, h, h2, h4, p, span } from 'react-hyperscript-helpers'
+import { a, div, h, h2, span } from 'react-hyperscript-helpers'
 import { ButtonPrimary, Link, Navbar, Select } from 'src/components/common'
+import HelpfulLinksBox from 'src/components/HelpfulLinksBox'
 import { centeredSpinner, icon } from 'src/components/icons'
 import { TextArea, TextInput } from 'src/components/input'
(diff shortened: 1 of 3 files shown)


src/pages/SubmissionConfig.js
```diff
@@ -245,115 +245,146 @@ export const SubmissionConfig = ({ methodId }) => {
...
+          p([
+            h(Link, { href: 'https://support.terra.bio/hc/en-us/articles/12029178977307-How-to-set-up-and-run-a-workflow', ...Utils.newTabLinkProps },
+              [
+                'How to set up and run a workflow',
+                icon('pop-out', { size: 12, style: { marginLeft: '0.25rem' } })
+              ]
+            )
+          ])
+        ])
+      ])
```
salonishah11, line 387: L359-L387 is the main change in this file. The changes above these lines are because I needed to wrap the workflow details in a separate div to create flex columns.

mspector: thanks for the callout!

cjllanwarne (approved):

src/pages/SubmissionConfig.js
mspector: this actually feels like it could be its own component... even if it's a small one. What do you think?

salonishah11: The links box has been refactored into a component.

mspector (approved):
nice!!

PR #139, opened 2023-03-29
[WM-1779] Add back actions button

kpierre13:


src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -122,6 +122,17 @@ export const SubmissionHistory = () => {
                 paddingTop: '1em'
               }),
               columns: [
+                {
```
salonishah11, line 125: We should hide the Actions column behind a feature flag. We don't want users to use this functionality until everything in Abort epic is complete.

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -122,6 +122,17 @@ export const SubmissionHistory = () => {
...
+                  size: { basis: 100, grow: 0 },
+                  field: 'actions',
+                  headerRenderer: () => h(Sortable, { sort, field: 'actions', onSort: setSort }, ['Actions']),
+                  cellRenderer: () => {
+                    return div(
+                      { style: { textAlign: 'center' } },
+                      [icon('cardMenuIcon', { size: 24, onClick: () => { window.alert('TODO: go to actions menu') } })]
```
salonishah11, line 132: For my own clarification - Is adding the `Abort Submission` option that does nothing part of this ticket or the next one? 🤔

salonishah11 (changes requested):

src/pages/SubmissionHistory/SubmissionHistory.js
kpierre13: Next ticket!

kpierre13: If everyone is alright with it, I think I'd prefer to do what @mspector suggested. With the code that has been changed as part of this ticket, it's a bit more difficult to incorporate the flag in a way that doesn't break the existing table, and I've already started working on the second part of this ticket, so it'd be much easier to use the flag in the followup since I'm changing this cell significantly. If it's alright, I can add in the logic that is adding a feature flag in this PR, and then use it in my next PR. What do you think @salonishah11 @mspector ?

salonishah11: I think the purpose of feature flag is to help us merge features that are still in development but aren't ready for the users to use yet. So I think this is a great opportunity to add a feature flag. The reason I am hesitant for not merging is that there are few more tickets for the abort functionality to be fully available to the user and having the code merged to dev with feature flag will also allow everyone to test the functionality sooner rather than later.

salonishah11: I think we should try and add the feature flag in this ticket so that it can be merged. But if it is non-trivial to add this flag, then I am fine wit doing what Michael suggested.

mspector (approved):

PR #141, opened 2023-03-30
[WM-1814] ajax contract tests

mspector:


.github/workflows/setup-and-test.yml
```diff
@@ -8,11 +8,14 @@ on:
     branches: [ '**' ]
   merge_group:
     branches: [ '**' ]
+  workflow_dispatch:
```
kpierre13, line 11: Is this a placeholder?

kpierre13 (approved):

.github/workflows/setup-and-test.yml
mspector: no! the presence of `workflow_dispatch:` without any arguments simply makes it possible to run manually from the web UI.

.github/workflows/setup-and-test.yml
```diff
@@ -28,6 +31,38 @@ jobs:
...
+        id: git-sha
+        run: |
+          CONSUMER_VERSION=$(git rev-parse --abbrev-ref HEAD)
+          echo "repo-branch=${CONSUMER_VERSION}" >> $GITHUB_OUTPUT
+          echo $CONSUMER_VERSION
+
+      - name: Encode the pact as non-breaking base 64 string
+        id: encode
+        env:
+          PACT_FULL_PATH: 'pacts/cbas-ui-cbas.json'  # this is a temporary hack until WM-1858 is done.
```
salonishah11, line 44: It might be good idea to create a ticket to update this when WM-1858 is done (unless it already exists then please ignore my comment)

src/libs/ajax-fetch.js
```diff
@@ -0,0 +1,47 @@
...
+}
+
+const withUrlPrefix = _.curry((prefix, wrappedFetch) => (path, ...args) => {
+  return wrappedFetch(prefix + path, ...args)
+})
+
+export const fetchOk = _.flow(withInstrumentation, withCancellation, withErrorRejection)(fetch)
+export const fetchCbas = withUrlPrefix(`${getConfig().cbasUrlRoot}/api/batch/v1/`, fetchOk)
+export const fetchCromwell = withUrlPrefix(`${getConfig().cromwellUrlRoot}/api/workflows/v1/`, fetchOk)
+export const fetchLeo = withUrlPrefix(`${getConfig().leoUrlRoot}/`, fetchOk) // TODO: How to add this config to Cromwhelm?
```
salonishah11, line 46: re TODO: I think `leoUrlRoot` should be set in Cromwhelm charts as part of [WM-1803](https://broadworkbench.atlassian.net/browse/WM-1803)

salonishah11 (approved):
Small nits otherwise LGTM! 👍

PR #142, opened 2023-04-03
[WM-1780] actions icon calls abort endpoint in cbas

kpierre13:
Abort icon appears with the option to abort a submission
image

Success notification when submitting request
image

### DIFF
--- a/src/pages/SubmissionHistory/SubmissionHistory.js
+++ b/src/pages/SubmissionHistory/SubmissionHistory.js
@@ -2,14 +2,17 @@ import _ from 'lodash/fp'
 import { Fragment, useRef, useState } from 'react'
 import { div, h, h2 } from 'react-hyperscript-helpers'
 import { AutoSizer } from 'react-virtualized'
-import { ButtonOutline, Link, Navbar } from 'src/components/common'
+import { ButtonOutline, Clickable, Link, Navbar } from 'src/components/common'
 import { centeredSpinner, icon } from 'src/components/icons'
+import { MenuButton, MenuTrigger } from 'src/components/PopupTrigger'
 import { AutoRefreshInterval, getDuration, isRunSetInTerminalState, loadAllRunSets, makeStatusLine, statusType } from 'src/components/submission-common'
 import { FlexTable, paginator, Sortable, tableHeight, TextCell } from 'src/components/table'
+import { Ajax } from 'src/libs/ajax'
 import colors from 'src/libs/colors'
 import * as Nav from 'src/libs/nav'
 import { notify } from 'src/libs/notifications'
 import { useCancellation, useOnMount } from 'src/libs/react-utils'
+import { maybeParseJSON } from 'src/libs/utils'
 import * as Utils from 'src/libs/utils'
 
 
@@ -40,6 +43,14 @@ export const SubmissionHistory = () => {
       notify('error', 'Error loading previous run sets', { detail: await (error instanceof Response ? error.text() : error) })
     }
   })
+
+  const cancelRunSet = async (submissionId) => {
+    try {
+      await Ajax(signal).Cbas.runSets.cancel(submissionId)
+    } catch (error) {
+      notify('error', 'Error canceling run set', { detail: await (error instanceof Response ? error.text() : error) })
+    }
+  }
 
   useOnMount(async () => {
     await refresh()
@@ -126,11 +137,19 @@ export const SubmissionHistory = () => {
                   size: { basis: 100, grow: 0 },
                   field: 'actions',
                   headerRenderer: () => h(Sortable, { sort, field: 'actions', onSort: setSort }, ['Actions']),
-                  cellRenderer: () => {
-                    return div(
-                      { style: { textAlign: 'center' } },
-                      [icon('cardMenuIcon', { size: 24, onClick: () => { window.alert('TODO: go to actions menu') } })]
-                    )
+                  cellRenderer: ({ rowIndex }) => {
+                    return h(MenuTrigger, {
+                      closeOnClick: true,
+                      'aria-label': 'Action selection menu',
+                      content: h(Fragment, [
+                        h(MenuButton, { /*onClick: () => { cancelRunSet(paginatedPreviousRunSets[rowIndex].run_set_id).then(r => pass)}*/}, ['Abort'])
+                      ]),
+                      style: { textAlign: 'center' }
+                      }, [
+                        h(Clickable, {
+                          'aria-label': 'Action selection menu',
+                        }, [icon('cardMenuIcon')])
+                      ])
                   }
                 },
                 {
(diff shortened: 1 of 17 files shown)


src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -123,16 +134,26 @@ export const SubmissionHistory = () => {
...
-                        [icon('cardMenuIcon', { size: 24, onClick: () => { window.alert('TODO: go to actions menu') } })]
-                      )
+                    headerRenderer: () => h(TextCell, {}, ['Actions']),
+                    cellRenderer: ({ rowIndex }) => {
+                      return h(MenuTrigger, {
+                        'aria-label': 'Action selection menu',
+                        content: h(Fragment, [
+                          h(MenuButton, {
+                            style: { fontSize: 15 },
+                            onClick: () => { cancelRunSet(paginatedPreviousRunSets[rowIndex].run_set_id) }
```
salonishah11, line 148: Will this allow user to abort a submission even if it is in a terminal state?

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -123,16 +134,26 @@ export const SubmissionHistory = () => {
...
+                          }, ['Abort'])
+                        ])
+                      }, [
+                        h(Clickable, {
+                          style: { textAlign: 'center' },
+                          'aria-label': 'Action selection menu'
+                        }, [icon('cardMenuIcon', { size: 35 })])
```
salonishah11, line 155: I am wondering if it would be good to reduce the spacing between the menu and the actions button? Based on the screenshot it feels that they are a bit detached. What do you think?

src/pages/SubmissionHistory/SubmissionHistory.js
kpierre13: I think so, although nothing happens. I'll add a check in `cancelRunSet` that'll notify a user that they cannot cancel a terminal run set.

src/pages/SubmissionHistory/SubmissionHistory.js
kpierre13: I think that will look better visually.

kpierre13:
Menu options are now close to the menu trigger:
image

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -123,16 +134,35 @@ export const SubmissionHistory = () => {
...
+                        'aria-label': 'Action selection menu',
+                        popupProps: {
+                          style: { left: '-20px' }
+                        },
+                        content: h(Fragment, [
+                          h(MenuButton, {
+                            style: { fontSize: 15 },
+                            onClick: () => {
+                              if (isRunSetInTerminalState(paginatedPreviousRunSets[rowIndex].state) === true) {
+                                notify('error', 'Error aborting run set', { detail: 'Cannot abort a submission that is in a terminal state.' })
```
salonishah11, line 153: I think instead of showing the error we should disable the Abort button with a tooltip saying something along the lines of `...this submission is already in terminal state`. This would align it with what we have in Terra UI

salonishah11 (approved):

src/pages/SubmissionHistory/SubmissionHistory.js
salonishah11: You might be able to use the same `isRunSetInTerminalState` for disabling the button

kpierre13: I forgot you gave an example error message here, but let me know if the one I have in the revision is good enough.

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -148,12 +149,12 @@ export const SubmissionHistory = () => {
                         content: h(Fragment, [
                           h(MenuButton, {
                             style: { fontSize: 15 },
+                            disabled: isRunSetInTerminalState(paginatedPreviousRunSets[rowIndex].state),
```
salonishah11, line 152: 🎉

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -46,8 +46,9 @@ export const SubmissionHistory = () => {
 
   const cancelRunSet = async submissionId => {
     try {
-      await Ajax(signal).Cbas.runSets.cancel(submissionId)
-      notify('success', 'Abort submission request submitted successfully', { message: 'You may refresh the page to get most recent status changes.', timeout: 5000 })
+      if (isRunSetInTerminalState(submissionId.state) === false) {
```
salonishah11, line 49: I think this condition would be redundant here because the user won't be able to reach here is the button is already disabled

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -123,16 +135,35 @@ export const SubmissionHistory = () => {
...
+                          style: { left: '-20px' }
+                        },
+                        content: h(Fragment, [
+                          h(MenuButton, {
+                            style: { fontSize: 15 },
+                            disabled: isRunSetInTerminalState(paginatedPreviousRunSets[rowIndex].state),
+                            tooltip: isRunSetInTerminalState(paginatedPreviousRunSets[rowIndex].state) && 'Cannot abort a terminal submission',
+                            onClick: () => {
+                              cancelRunSet(paginatedPreviousRunSets[rowIndex].run_set_id)
+                              notify('success', 'Abort submission request submitted successfully',
```
salonishah11, line 156: This notify should be moved inside `cancelRunSet` inside the try block otherwise I think this notify will always be displayed irrespective of whether abort was success or not 🤔

salonishah11 (commented):
Thank you for implementing the feedback. Small nits otherwise this looks great! 👍

src/pages/SubmissionHistory/SubmissionHistory.js
kpierre13: Yes, I meant to get rid of this, thank you for catching!

src/pages/SubmissionHistory/SubmissionHistory.js
kpierre13: Hm, I see what you mean. I had the success inside of `cancelRunSet` initially, but this caused a problem where when the page is rendered, there would be constant X amount of success messages depending on how many submissions a user had. This is weird considering the success is inside of an `onClick`, so I don't know why we get the notification on a refresh or initial page render.

kpierre13: Ok, I think I resolved this. I'm not sure why I was seeing what I was seeing. I also added an additional check for the button being disabled when the state is also `Canceling`.

mspector (approved):

PR #144, opened 2023-04-10
[WM-1857] Support WDS as a standalone app

kpierre13:


src/components/submission-common.js
```diff
@@ -107,25 +106,30 @@ export const resolveWdsApp = apps => {
...
 
-  //Failed to find an app with the proper name, look for a RUNNING WDS app
-  const runningWdsApps = apps.filter(app => app.appType === getConfig().wdsAppTypeName && app.status === 'RUNNING')
-  if (runningWdsApps.length > 0) {
-    // Evaluate the earliest-created WDS app
-    runningWdsApps.sort((a, b) => new Date(a.auditInfo.createdDate).valueOf() - new Date(b.auditInfo.createdDate).valueOf())
-    return runningWdsApps[0]
-  }
+  // WDS appType is checked first and takes precedence over CROMWELL apps in the workspace
+  const wdsAppTypes = ['WDS', 'CROMWELL']
```
salonishah11, line 111: The app type names should be fetched from config as before. It might be good to add a comment in [WM-1803](https://broadworkbench.atlassian.net/browse/WM-1803) mentioning that the `wdsAppTypeName` should be an array instead of single value. FYI @mspector since you are working on WM-1803

src/components/submission-common.test.js
```diff
@@ -85,4 +85,20 @@ describe('resolveWdsUrl', () => {
...
+  it.each(
+    [
+      { appStatus: 'RUNNING', expectedUrl: mockWdsProxyUrl },
+      { appStatus: 'PROVISIONING', expectedUrl: '' },
+      { appStatus: 'STOPPED', expectedUrl: '' },
+      { appStatus: 'STOPPING', expectedUrl: '' }
+    ]
+  )('gives precedence to the WDS appType over the CROMWELL appType', ({ appStatus, expectedUrl }) => {
+    const uuid = '123e4567-e89b-12d3-a456-426614174000'
+    const testHealthyAppProxyUrlResponse: Array<Object> = [
```
salonishah11, line 98: The type definition `Array<Object>` is needed for Typescript files only. We can remove this here.

kpierre13: Thank you for catching this!

src/components/submission-common.js
```diff
@@ -107,25 +107,29 @@ export const resolveWdsApp = apps => {
...
 
-  //Failed to find an app with the proper name, look for a RUNNING WDS app
-  const runningWdsApps = apps.filter(app => app.appType === getConfig().wdsAppTypeName && app.status === 'RUNNING')
-  if (runningWdsApps.length > 0) {
-    // Evaluate the earliest-created WDS app
-    runningWdsApps.sort((a, b) => new Date(a.auditInfo.createdDate).valueOf() - new Date(b.auditInfo.createdDate).valueOf())
-    return runningWdsApps[0]
-  }
+  // WDS appType is checked first and takes precedence over CROMWELL apps in the workspace
+  for (const appTypeName of getConfig().wdsAppTypeName) {
```
salonishah11, line 112: small nit: Since `wdsAppTypeName` is an array maybe we can rename it to `wdsAppTypeNames`? But I don't have a strong opinion on this either way.

salonishah11 (approved):

src/components/submission-common.js
mspector: I second this. I actually think it's kind of important!

mspector: I think I remember us talking about some logic that is being borrowed from elsewhere (maybe terra-ui). Is that logic? Just curious

mspector (approved):
LGTM! Though I do think it should be `wdsAppTypeNames`, plural

src/components/submission-common.js
kpierre13: Should this be renamed to something like `appTypeNames` since there will be multiple apps?

PR #151, opened 2023-04-21
[WM-1811] post run sets contract test

mspector:


.github/workflows/setup-and-test.yml
```diff
@@ -34,7 +34,7 @@ jobs:
       - name: Get the git SHA and use it as the repo-branch
         id: git-sha
         run: |
-          CONSUMER_VERSION=$(git rev-parse --abbrev-ref HEAD)
+          CONSUMER_VERSION=${{ github.event.pull_request.head.sha }}
```
salonishah11, line 37: For my own understanding: what does this change do?

src/libs/ajax.test.js
```diff
@@ -79,4 +88,169 @@ describe('Ajax tests', () => {
...
+
+    const body = JSON.stringify(payload)
+    const headers = { 'Content-Type': 'application/json' }
+
+    await cbasPact.addInteraction({
+      states: [
+        { description: 'ready to fetch recordId FOO1 from recordType FOO from wdsService' },
+        { description: 'ready to fetch myMethodVersion with UUID 90000000-0000-0000-0000-000000000009' },
+        { description: 'ready to receive exactly 1 call to POST run_sets' }
+      ],
```
salonishah11, line 177: Can you remind me again what is the purpose of these states and descriptions? It seems it is being repeated in all 3 tests that are added in this file

mspector: Each interaction is allowed to specify any number of states that it needs for the interaction to work as intended. In this case, because we want to be able to call `POST run_sets`, we need there to be (1) records in the record table, (2) a methodVersion, and (3) to be able to verify the run set ID of the run set that will be returned in this interaction.

When CBAS runs its verification tests against these interactions, these states will tell it how to set itself up, using `@State` functions that match each of the above descriptions exactly.

The reason each of the tests has the same set of states is that they're all testing `POST run_sets` functionality - the same set of three states are needed to test each of the different `POST run_sets` calls that these tests run.

That we can reuse these states for many tests is a really good thing, because it means we don't need to have a 1:1 relationship between consumer tests and provider tests. The provider can just set up a bunch of states that many consumer tests can use multiple times.

salonishah11 (approved):

kpierre13 (approved):

PR #152, opened 2023-04-24
Add Workspace Manager values to config.json

THWiseman:


public/config.json
```diff
@@ -2,5 +2,8 @@
   "cbasUrlRoot" : "http://localhost:3000",
   "cromwellUrlRoot" : "http://localhost:3000/cromwell",
   "wdsUrlRoot" : "http://localhost:3000/wds",
-  "wdsInstanceId": ""
+  "wdsInstanceId" : "",
+  "workspaceManagerRoot" : "",
+  "workspaceId" : "",
```
salonishah11, line 7: My assumption was that `wdsInstanceId` is same as `workspaceId`. Are they different?

THWiseman: You're correct! They're the same now (and coming from the same helm value), but I believe they might diverge in the future. I figure its best that stuff that talks to WDS use `wdsInstanceId`  and stuff that talks to WSM use `workspaceId` in case they do end up splitting. Happy to change if that doesn't make sense or I'm misinformed.

salonishah11: I see. I had assumed that even in multi-user CBAS land, since we still would have 1 CBAS & WDS per workspace `wdsInstanceId` and `workspaceId` would remain same. But I guess depending on how far in the future they might diverge, maybe it would be better to keep them the same until the need arises? What do you think?

salonishah11: Another thought I had is to maybe rename `wdsInstanceId` to `workspaceId` since the former is actually referring to workspace ID and calling it `wdsInstanceId` leads to confusion. I am interested to see what you and other reviewers think about this thought.

salonishah11: Discussed face to face and it was decided that keeping `wdsInstanceId` and `workspaceId ` separate seems the right path forward.

salonishah11 (approved):

jgainerdewar (approved):

PR #153, opened 2023-04-25
[WM-1859] BYO URL v0

kpierre13:
When the flag is enabled `configOverridesStore.set({ isFindWorkflowEnabled: true, isURLEnabled: true })` the url workflow link shows in the sidebar and users are able to post a url with a workflow name and version.
image


src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -128,9 +154,46 @@ const FindWorkflowModal = ({ onDismiss }) => {
       ]),
       isSubHeaderActive('browse-suggested-workflows') && div({ style: { overflowY: 'auto', flexGrow: 1, display: 'flex', flexDirection: 'column', paddingLeft: '20px' } }, [
         div({ style: { display: 'flex', flexWrap: 'wrap', overflowY: 'auto', paddingBottom: 5, paddingLeft: 5 } }, [
-          _.map(method => h(MethodCard, { method, onClick: () => submitMethod({ method }), key: method.method_name }), suggestedWorkflowsList)
+          _.map(method => h(MethodCard, { method, onClick: () => submitMethod(method), key: method.method_name }), suggestedWorkflowsList)
         ])
-      ])
+      ]),
+      getConfig().isURLEnabled && isSubHeaderActive('add-a-workflow-link') && div({ style: { marginLeft: '4rem' } }, [
```
cjllanwarne, line 160: Presumably the subHeader becoming active is only possible if the url enabled is already true? I might be tempted to simplify this condition by basing it only on the subheader, just like the other "browse suggested workflows" option.

src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -128,9 +154,46 @@ const FindWorkflowModal = ({ onDismiss }) => {
...
+        div({ style: { width: 500 } }, [h2(['Workflow Link'])]),
```
cjllanwarne, line 161: TOL: This file is just on the edge of becoming too long to understand all in one go. Rather than adding this all inline, might it make sense to have a `ImportGithub` class that can (1) keep the code complexity in a separate file and (2) be tested as a unit, instead of as part of a larger `FindWorkflowModal` component?

src/components/HelpfulLinksBox.js
```diff
@@ -9,15 +9,15 @@ import * as Utils from 'src/libs/utils'
 const HelpfulLinksBox = ({ method }) => {
   return div({ style: { backgroundColor: colors.accent(0.2), paddingTop: '0.25em', paddingBottom: '0.25em', paddingLeft: '1em', paddingRight: '1em' } }, [
     h4('Have questions?'),
-    isCovid19Method(method.name) && p([
+    isCovid19Method(method?.name) ? p([
```
salonishah11, line 12: I think`?` can be reverted to `&&` and it should still work as expected. This way `: null` on L19 won't be necessary

src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -91,8 +100,25 @@ const FindWorkflowModal = ({ onDismiss }) => {
     }
   })
 
+  const reconstructToRawUrl = url => {
```
salonishah11, line 103: It might be good to add unit tests for this method

src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -67,19 +70,25 @@ const suggestedWorkflowsList = [
...
+  const [workflowName, setWorkflowName] = useState('')
+  const [versionName, setVersionName] = useState('')
+  const [methodUrl, setMethodUrl] = useState('')
 
   const signal = useCancellation()
 
-  const submitMethod = withBusyState(setLoading, async ({ method }) => {
+  const submitMethod = withBusyState(setLoading, async method => {
     try {
+      const rawGithubUrl = reconstructToRawUrl(method.method_url)
```
salonishah11, line 81: To clarify my understanding: what happens if the `method_url` is already a raw github url? 🤔

src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -67,19 +70,25 @@ const suggestedWorkflowsList = [
 const FindWorkflowModal = ({ onDismiss }) => {
   const [selectedSubHeader, setSelectedSubHeader] = useState('browse-suggested-workflows')
   const [loading, setLoading] = useState()
+  const [workflowName, setWorkflowName] = useState('')
+  const [versionName, setVersionName] = useState('')
```
salonishah11, line 74: small nit: since all these are method related I am thinking it might be better to name them as `methodName` and `methodVersionName`. What do you think?

src/pages/FindWorkflow/FindWorkflowModal.js
kpierre13: It still works. My understanding of the find and replace is that if it finds the substring, it'll replace it, if not, nothing will happen and the url remains unchanged.

src/components/ImportGithub.js
```diff
@@ -0,0 +1,60 @@
+import { Fragment, useState } from 'react'
+import { div, h, h2 } from 'react-hyperscript-helpers'
+import { ButtonPrimary } from 'src/components/common'
+import { TextInput } from 'src/components/input'
+import { submitMethod } from 'src/components/method-common'
+import { withBusyState } from 'src/libs/utils'
+
+// { tooltip: 'Link must start with https://github.com or https://raw.githubusercontent.com' },
```
cjllanwarne, line 8: Is this meant to be commented out up here?

salonishah11, line 8: Should we add this is as a validation on workflow url field?

cjllanwarne (approved):
💯

src/components/ImportGithub.js
```diff
@@ -0,0 +1,60 @@
...
+      h(TextInput, {
+        style: { width: 500 },
+        placeholder: 'Paste Github Link',
+        value: methodUrl,
+        onChange: u => setMethodUrl(u),
+        'aria-label': 'Github link input'
+      })
+    ]),
+    div({ style: { marginTop: '3rem', width: 500 } }, [
+      h2(['New Workflow Name / Version'])
```
salonishah11, line 26: small nit: I feel the `h2` is making the input names appear too big and out of place in context to the sub-headers `Browse Suggested Workflows` and `Add a Workflow link` and the text size in Helpful Hints box. Maybe `h3` would be a better heading format here? What do you think?
h2:
Screenshot 2023-05-05 at 11 10 17 AM
h3:
Screenshot 2023-05-05 at 11 10 58 AM

src/components/ImportGithub.js
```diff
@@ -0,0 +1,60 @@
...
+import { withBusyState } from 'src/libs/utils'
+
+// { tooltip: 'Link must start with https://github.com or https://raw.githubusercontent.com' },
+const ImportGithub = ({ setLoading, signal, onDismiss }) => {
+  const [methodName, setMethodName] = useState('')
+  const [methodVersionName, setMethodVersionName] = useState('')
+  const [methodUrl, setMethodUrl] = useState('')
+
+  return div({ style: { marginLeft: '4rem' } }, [
+    div({ style: { width: 500 } }, [h2(['Workflow Link'])]),
```
salonishah11, line 15: TOL: Should we put red `*` or the orange warning `!` next to each field names to indicate that all 3 fields are required?

src/components/ImportGithub.js
kpierre13: Ah, thanks for catching. I meant to ask it as a question if we should add an info icon with a tooltip? I think it would be beneficial to users so they know what a valid link looks like. What do you think?

kpierre13: I meant to pose the question to you and Chris. I think it could be useful to have in case there's a question about what a proper link looks like.

src/components/ImportGithub.js
kpierre13: If I disable the button until all three fields have entries, would it be too much to also have 3 warnings as well? Should we do one or the other?

src/components/ImportGithub.js
salonishah11: I think that's a good idea. We can add this tooltip. But I also think we should validate the url entered before sending it to CBAS. Validating in the UI will help avoid unnecessary trips to the API just for the import to fail. If the url is invalid the user will be able to edit the url in UI itself.

src/components/ImportGithub.js
salonishah11: Yeah I think it is worth adding either the red `*` or orange `!` next to these fields so that user knows they are required (just like we do with inputs). And disabling the `Add` button until all 3 fields have values will ensure we don't allow sending empty values to API (where it will eventually fail)

salonishah11: I was trying to find if we have the notion of `*` in Terra UI and it seems we do have a black colored `*`. And the create button is disabled with error stating why it is disabled. 
Screenshot 2023-05-08 at 11 16 35 AM

cjllanwarne (approved):

salonishah11 (approved):
💯

PR #159, opened 2023-05-17
[WM-1980] Dockstore Library option in Find Workflow Modal

kpierre13:


src/libs/utils.js
```diff
@@ -189,3 +189,15 @@ export const renderTypeText = iotype => {
   }
   return 'Unsupported Type'
 }
+
+export const getDockstoreUrlRoot = () => {
+  let rootUrl
+  switch (process.env.NODE_ENV) {
```
salonishah11, line 195: Should Leonardo also set the Dockstore url in cbas-ui's cromwhelm chart instead of getting it from `process` just as we discussed in today's standup? 🤔

kpierre13: I think if that's what you're doing from your ticket, then that makes a lot of sense for consistency. How would this change when integrating back into Terra-UI?

salonishah11: Yeah it would be a more correct approach and consistent too. I think those changes should be part of this ticket to ensure the feature works as expected. I think when it is reintegrated into Terra UI, the logic would change as Terra UI already has logic in place to get Dockstore root url based on env as Terra UI does know which env the app is running in.

src/libs/config.js
```diff
@@ -14,3 +14,4 @@ export const getConfig = () => {
 export const isFindWorkflowEnabled = () => getConfig().isFindWorkflowEnabled
 export const isActionMenuEnabled = () => getConfig().isActionMenuEnabled
 export const isURLEnabled = () => getConfig().isURLEnabled
+export const isDockstoreEnabled = () => getConfig().isURLEnabled
```
dillydally414, line 17: Is this meant to be the same as isURLEnabled?

dillydally414 (approved):

mspector (approved):

PR #160, opened 2023-05-23
[WM-1940] Remove Abort feature flag

kpierre13:


salonishah11 (approved):

aednichols (approved):

src/pages/SubmissionHistory/SubmissionHistory.test.js
```diff
@@ -349,22 +343,6 @@ describe('SubmissionHistory page', () => {
...
     expect(screen.getByText('Some submission statuses are not up to date. Refreshing the page may update more statuses.')).toBeInTheDocument()
   })
-})
-
-describe('Actions column', () => {
-  beforeEach(() => {
-    getConfig.mockReturnValue({ isActionMenuEnabled: true })
-  })
-
-  it('Shows Actions column on table', async () => {
```
cjllanwarne, line 359: This looks like it should still be valid? In fact... even more valid now, since it should _always_ be shown?

src/pages/SubmissionHistory/SubmissionHistory.js
```diff
@@ -134,35 +133,32 @@ export const SubmissionHistory = () => {
                 paddingTop: '1em'
               }),
               columns: [
-                ...getConfig().isActionMenuEnabled ?
```
cjllanwarne, line 137: If a submission is finished, do we still want an empty action menu next to all the workflows?

src/pages/SubmissionHistory/SubmissionHistory.test.js
kpierre13: I figured the other tests are now looking for 6 columns, so that should cover checking for its existence. No need for a specific test.

src/pages/SubmissionHistory/SubmissionHistory.js
kpierre13: Are you suggesting that the action column disappear for a finished submission? Currently, the only action is `Abort` and is disabled for terminal submissions. Do we think there are other Actions that may be part of this column in the future?

PR #169, opened 2023-05-31
[WM-1906] Optional Input Filtering

dillydally414:
Screenshot 2023-06-01 at 10 33 38 AM
Screenshot 2023-06-01 at 10 33 54 AM
Default sort: attribute name primary, variable name secondary.


src/components/InputsTable.js
```diff
@@ -119,12 +125,22 @@ const InputsTable = props => {
           setStructBuilderVisible(false)
         }
       }),
+      (!includeOptionalInputs || _.some(row => row.optional, inputTableData)) && h(div,
+        { style: { height: 0.08 * height, width, display: 'flex', alignItems: 'center' } },
```
mspector, line 129: here's that same style definition again!

mspector (approved):
LGTM, except for the minor nit about storing the style definition in a variable. I'm not adamant about that, but I have seen it reused twice now

mspector (approved):
LGTM, except for the minor nit about storing the style definition in a variable. I'm not adamant about that, but I have seen it reused twice now

salonishah11:
@dillydally414 this is great. Can we also add the optional inputs filtering to the [Struct modal](https://github.com/DataBiosphere/cbas-ui/blob/main/src/components/StructBuilder.js) as well?

dillydally414:
@salonishah11 I will look into that tomorrow - I might need some help finding an example workflow with a struct as I haven't ran into that yet.

salonishah11:
@dillydally414 thank you! I think the input `sample_and_unmapped_bams` in `ExomeGermlineSingleSample` has both required and optional inputs. You can probably use that for testing. Let me know if you need any help finding that workflow and I will be happy to pair with you on it.

src/components/InputsTable.js
dillydally414: Where would you suggest saving the style? I looked at the styles file but it doesn't look like any styles from there are used in this page. Also, this will remain as one style definition once the PRs are merged, since this style is just for the button row above the table and both buttons will be in the same row. I may abstract out the logic for showing this row into a helper function for readability once one of the PRs is merged, to make it clearer what this row is and when it's shown.

src/components/table.js
```diff
@@ -262,6 +262,18 @@ const NoContentRow = ({ noContentMessage, noContentRenderer = _.noop, numColumns
...
 ])
 
+export const InputsButtonRow = ({ showRow = true, optionalButtonProps: { includeOptionalInputs, setIncludeOptionalInputs }, ...props }) => {
+  return showRow && h(div, { ...props }, [
+    h(Link,
+      {
+        style: { marginRight: 'auto' },
+        onClick: () => setIncludeOptionalInputs(includeOptionalInputs => !includeOptionalInputs)
+      },
+      [includeOptionalInputs ? 'Hide optional inputs' : 'Show optional inputs']
```
salonishah11, line 272: TOL: I wonder if it would be better to right align this link? I guess I am thinking that since it is related to Attributes column, I mentally picture this on the right side of table where the Attributes column is. But I don't have a strong preference.

salonishah11 (approved):
small nit otherwise LGTM

src/components/table.js
dillydally414: Screenshot 2023-06-06 at 10 32 05 AM

I took the left-align style from Terra UI, screenshot above. Some of the other tickets either currently on my plate or upcoming will involve adding new things to that top bar ([1962](https://broadworkbench.atlassian.net/browse/WM-1962), [1949](https://broadworkbench.atlassian.net/browse/WM-1949)), so I'm definitely open to having a conversation at some point regarding the layout of the top bar.

PR #171, opened 2023-06-01
[WM-1962] Recommend Inputs from Data Table

dillydally414:
Screenshot 2023-06-01 at 1 56 11 PM
Screenshot 2023-06-01 at 1 56 30 PM

(See base_q_threshold)

Clicking on link auto populates attribute name as well as fetch from data table. Button in top right does this for all variables that have a column with the same name in the currently selected data table.

UPDATED SCREEN RECORDING BELOW

https://github.com/DataBiosphere/cbas-ui/assets/67511512/77a3bc75-3797-4dc7-919d-ead93dda0d6a


src/pages/SubmissionConfig.test.js
```diff
@@ -1307,6 +1307,150 @@ describe('SubmissionConfig inputs/outputs definitions', () => {
...
+    const cells2 = within(rows[2]).queryAllByRole('cell')
+    const cells3 = within(rows[3]).queryAllByRole('cell')
+
+    within(cells1[0]).getByText('foo')
+    within(cells1[1]).getByText('foo_rating')
+    within(cells1[2]).getByText('Int')
+    within(cells1[3]).getByText('None')
+    within(cells1[4]).getByText(/Use /)
+    const inputFillButton = within(cells1[4]).getByText('foo_rating')
+    within(cells1[4]).getByText(/ from data table?/)
```
mspector, line 1361: I was wondering about this - what does it look like when a required attribute is present in the data table? Would you mind taking a screenshot?

salonishah11:
@dillydally414 for some reason when I run this code locally I am not seeing the changes in the Inputs table UI. Am I doing something wrong? I have your branch checked out and running it locally. Do I need to do something else to test this? 🤔

dillydally414:
Do you have a data table locally with columns matching the names of variables? This suggestion only shows up in that case, so I had to create a new table in order to effectively test this. I'm heading out for the day but can follow up tomorrow to pass along the database details.

src/components/InputsTable.js
```diff
@@ -119,12 +126,25 @@ const InputsTable = props => {
...
+        { style: { height: 0.08 * height, width, display: 'flex', alignItems: 'center' } },
+        [h(Link,
+          {
+            style: { marginLeft: 'auto' },
+            onClick: () => _.flow(
+              _.filter(row => _.has(row.variable, dataTableAttributes)),
+              _.forEach(row => setConfiguredInputDefinition(_.set(`[${row.configurationIndex}].source`, { type: 'record_lookup', record_attribute: row.variable })))
+            )(inputTableData)
+          },
+          ['Set all from data table']
```
cjllanwarne, line 139: "all" is a slight misnomer. Can we maybe count how many we're proposing to auto-fill and say `Set (5) from data table`? (and maybe with a tooltip telling us exactly which 5 we found?)

dillydally414: Screenshot 2023-06-02 at 11 04 52 AM

Something to note, this tooltip does have a somewhat strange output since some variables are part of different tasks but have the same name

src/pages/SubmissionConfig.test.js
dillydally414: Screenshot 2023-06-02 at 11 09 15 AM

Same text content, not italic since not optional, and maintains the ⚠️ This input is required warning. I can change this to a required input in the tests to make sure it still works.

src/components/InputsTable.js
salonishah11: Two thoughts on the tooltip:
- I think if there are lot of inputs whose attribute values can be auto-filled, this tooltip would get verbose and long. I like the idea of tooltip but I am not sure how to best display the information. I am not sure if its possible, but can a tooltip have scroll bar?
- I think we should display the entire input name including the Task name so that variables with same names can be displayed. For example `ExomeGermlineSingleSample. sample_and_unmapped_bams`

dillydally414: Screenshot 2023-06-06 at 6 26 26 PM

Minor styling update

dillydally414:
> @dillydally414 do you know if this auto-filling of inputs can also be added to Struct builder component easily? If possible it would be great to add this option there also.

I discussed this during the walkthrough with the rest of the team and we determined this is out of the scope of this ticket - there would likely be a lot of refinement required to narrow down how columns are recommended, how the "set X" is displayed, etc.

mspector (approved):
thanks for the walkthrough! 👍 

We talked about getting Katie's feedback on this, but just to get it down in writing: the text "Use $attribute from data table?" is, in my opinion, too visually similar to other kinds of messages that appear in that column. I thought the suggestion to use an icon of some kind (even a cheesy lightbulb) was a good one.

I'd also mildly prefer a phrase that isn't a question - my suggestion would be "Autofill $attribute from data table", and to use "Autofill <rows> from data table" - but I'm really just nitpicking.

Neither of these suggestions block approval for me - I defer to Katie, Sara, and the rest of the team on both of these points

kpierre13 (approved):

PR #173, opened 2023-06-06
[WX-1143] Input/Output Modal

THWiseman:
Added a modal to the Call Table that allows users to view the Inputs and Outputs of a particular task. 
If a particular input or output is a blob file, the user can download the file directly by clicking on it. 

Tested locally and deployed to my dev environment. 

image


src/pages/RunDetails.test.js
```diff
@@ -181,6 +183,10 @@ describe('RunDetails - render smoke test', () => {
         expect(stdout).toBeDefined
         const stderr = within(row).getByText('stderr')
         expect(stderr).toBeDefined
+        const inputs = within(row).getByText('Inputs')
+        expect(inputs).toBeDefined
+        const outputs = within(row).getByText('Outputs')
+        expect(outputs).toBeDefined
```
aednichols, line 189: This kind of testing seems sensitive to any kind of localization or copy changes and thus seems too closely coupled  (for my personal subjective standard)

aednichols (approved):

src/pages/RunDetails.test.js
THWiseman: Makes sense to me! I'll update a few tests to use testids instead.

jgainerdewar (approved):

PR #174, opened 2023-06-06
[WX-1147] Fix File Not Found Error template on URIViewer

JVThomas:
Addresses [WX-1147](https://broadworkbench.atlassian.net/browse/WX-1147)

PR fixes error message rendering on `URIViewer's` error message template when it fails to retrieve a file. This functionality is required in situations where `stdout` and `stderr` are not available because Cromwell was never able to start a task to begin with (usually due to errors like invalid/accessible docker images and/or inputs).

PR also adds tests to check against the expected functionality.

[WX-1147]: https://broadworkbench.atlassian.net/browse/WX-1147?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/libs/ajax.js
+++ b/src/libs/ajax.js
@@ -152,6 +152,12 @@ const AzureStorage = signal => ({
     const sasToken = await WorkspaceManager(signal).getSASToken()
     const url = `${blobFilepath}?${sasToken}`
     const res = await fetchAzureStorage(url, _.mergeAll([{ signal, method: 'GET' }]))
+    // Terra-UI and CBAS-UI use the Fetch API which treats any network response that returns as a resolved promise
+    // This is in contrast to other packages (ex: Axios) which treat any network response with a status code of 4xx or 5xx as a rejected promise
+    // Since the modal hinges on the file being returned, we'll treat any non-200 response as an error
+    if (res.status !== 200) {
+      throw new Error('Failed to fetch file from Azure Blob Storage')
+    }
     const textContent = await res.text()
     const blobDetails = parseAzureBlobUri(blobFilepath)
     const ret =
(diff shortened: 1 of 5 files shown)


THWiseman (approved):

src/libs/ajax.js
```diff
@@ -152,6 +152,12 @@ const AzureStorage = signal => ({
     const sasToken = await WorkspaceManager(signal).getSASToken()
     const url = `${blobFilepath}?${sasToken}`
     const res = await fetchAzureStorage(url, _.mergeAll([{ signal, method: 'GET' }]))
+    // Terra-UI and CBAS-UI use the Fetch API which treats any network response that returns as a resolved promise
+    // This is in contrast to other packages (ex: Axios) which treat any network response with a status code of 4xx or 5xx as a rejected promise
+    // Since the modal hinges on the file being returned, we'll treat any non-200 response as an error
+    if (res.status !== 200) {
+      throw new Error('Failed to fetch file from Azure Blob Storage')
```
aednichols, line 159: Is this code a bit brittle in case the API uses a different 2xx range success code under some conditions? I could see 204 perhaps.

aednichols (approved):
One 100% optional thought

src/libs/ajax.js
dillydally414: I could be wrong but doesn't the `fetchAzureStorage` implementation already throw an error if the status isn't in the ok range/2xx? Looking [here](https://github.com/DataBiosphere/cbas-ui/blob/main/src/libs/ajax-fetch.js#L30).

THWiseman: (speaking on behalf of Justin while he's off. We worked on this together yesterday.) 

Both of you make excellent points. I think for this function specifically, we're looking for exactly a 200 from Azure. We not only need the HTTP request to succeed, but we also need it to return a text file that we can display to the user. Yes, it's a little brittle, but we're asking it to do/return a very specific thing.

PR #175, opened 2023-06-07
[WM-1948] Literal Array Inputs

dillydally414:
Screenshot 2023-06-07 at 10 28 26 AM
Screenshot 2023-06-07 at 10 28 42 AM

Note that top error, unlike other errors, will not prevent the user from submitting - it is just there so they're aware there is likely an issue with their input, but we _can_ still parse it as a one-element array. Both of these inputs would be sent to CBAS in an identical manner.

UPDATED SCREEN RECORDING BELOW

https://github.com/DataBiosphere/cbas-ui/assets/67511512/2cf657d7-bd10-4f6f-b664-d80016bd17ac


src/components/InputsTable.js
```diff
@@ -69,7 +70,12 @@ const InputsTable = props => {
   const parameterValueSelectWithWarnings = (rowIndex, selectedInputName) => {
     const warningMessage = Utils.cond(
       [missingRequiredInputs.includes(selectedInputName), () => 'This attribute is required'],
-      [inputsWithInvalidValues.includes(selectedInputName), () => 'Value is either empty or doesn\'t match expected input type'],
+      [inputsWithInvalidValues.includes(selectedInputName) && inputTableData[rowIndex].source.parameter_value === '', () => 'Value is empty'],
```
cjllanwarne, line 73: Packing all of this into `inputsWithInvalidValues` seems to be having the side effects of:
1. Forcing you to override the "warning shape" and "warning color", in one specific situation, and
2. Reinterpreting the error messages later on to decide whether they mean we need to block submission or not

I wonder whether we could reframe this as "`inputsWithMessages`", a list of pairs/objects which look like `{messageType, message}`. Then:
1. It's obvious when we're rendering the message which icons/shapes we should use (and all errors/infos/validations/...) are all guaranteed to look the same
2. When we're checking whether any messages should block submission, we just need to check for "error" type messages. Or we could have a third `{isBlocker}` field on the objects.

kpierre13 (approved):
Great job!

src/components/submission-common.js
```diff
@@ -222,42 +229,46 @@ export const ParameterValueTextInput = props => {
...
 
+  const placeholder = Utils.cond(
+    [unwrapOptional(inputType).type === 'primitive' && unwrapOptional(inputType).primitive_type === 'String', () => '(Empty string)'],
+    [unwrapOptional(inputType).type === 'array', () => Utils.switchCase(unwrapOptional(unwrapOptional(inputType).array_type).primitive_type,
+      ['Int', () => '[1, 2, 3]'],
+      ['Float', () => '[1.5, 2.0, 5]'],
+      ['String', () => '["value1", "value2", "value3"]'],
+      ['Boolean', () => '[true, false, true]']
+    )],
+    () => undefined
```
cjllanwarne, line 254: This use of `undefined` to mean "all good" makes me a little nervous for one good and one less-good reason -
- My good reason: If something goes wrong in calculating the message, it's pretty likely that the function will return `undefined` by default and we'll accidentally assume that means "all good"
- My less-good reason: I have a general preference for strong typing and avoiding values like `null`s and `undefined`s as return values from functions, eg by having "active choice" values to mean "no value here".

I wonder what you think about one of these options:
1. Maybe it's a good idea to return a "success" value whenever we validate a value and it looks good? Might make people more inclined to look at what those info message are on slightly questionable inputs if everything else in their table is green?
2. Maybe we can return a "no comment" message type and map that into no change in the input table UI?

src/components/submission-common.test.js
```diff
@@ -269,6 +281,71 @@ describe('requiredInputsWithoutSource', () => {
...
+      arrayInput('validString', 'String', ['a', 'b', 'c']), // success
+      arrayInput('validStringArray', 'String', '["a", "b", "c"]'), // success
+      arrayInput('validStringSingleton', 'String', 'foo'), // info
+      arrayInput('invalidStringEmpty', 'String', '') // error
+    ]
+
+    const inputMessages = validateInputs(inputsWithArraysDefinition, emptyDataTableAttributes)
+    const errorInputs = _.flow(_.filter(message => message.type === 'error'), _.map(message => message.name))(inputMessages)
+    const infoInputs = _.flow(_.filter(message => message.type === 'info'), _.map(message => message.name))(inputMessages)
+    const successInputs = _.flow(_.filter(message => message.type === 'success'), _.map(message => message.name))(inputMessages)
```
cjllanwarne, line 339: TOL: If there were 2 types we could use `partition` to do all of this in a single pass... I wonder whether there's an equivalent way to do a 3-way split in a single pass over the initial list

src/components/submission-common.test.js
```diff
@@ -269,6 +281,71 @@ describe('requiredInputsWithoutSource', () => {
...
+      arrayInput('invalidBooleanString', 'Boolean', [true, 'a', false]), // error
+
+      // string arrays
+      arrayInput('validString', 'String', ['a', 'b', 'c']), // success
+      arrayInput('validStringArray', 'String', '["a", "b", "c"]'), // success
+      arrayInput('validStringSingleton', 'String', 'foo'), // info
+      arrayInput('invalidStringEmpty', 'String', '') // error
+    ]
+
+    const inputMessages = validateInputs(inputsWithArraysDefinition, emptyDataTableAttributes)
```
cjllanwarne, line 336: TOL: I wonder whether this is really `validations` now instead of `messages`? You could argue the main output is the set of enum values we use to decide "is this input invalid" (to disable the submit button), and that the messages are incidental. It might also feel better to have some kind of active "All good, no further comment" value if this is called "validations" instead of "messages"

src/pages/SubmissionConfig.test.js
```diff
@@ -969,7 +971,7 @@ describe('Input source and requirements validation', () => {
 
     // ** ASSERT **
     // check that tooltip indicating missing required fields is present for Submit button
-    screen.getByText('One or more inputs have missing values')
+    screen.getByText('One or more inputs have missing/invalid values')
```
cjllanwarne, line 974: Sorry to comment on tests instead of real code, but this is more of an "is this the outcome we want" style comment... Would it be hard to actually count the instances and make this be `"2 inputs have missing/invalid values"`

src/pages/SubmissionConfig.test.js
```diff
@@ -1074,23 +1076,179 @@ describe('Input source and requirements validation', () => {
 
     // ** ASSERT **
     // check that the warning message for input exists
-    within(firstInputRowCells[4]).getByText('This attribute is required')
+    within(firstInputRowCells[4]).getByText('Value is empty')
```
cjllanwarne, line 1079: 💯 !

cjllanwarne (commented):
A couple of mostly TOL-level comments, and if we can, please let's also include `File`s as acceptable primitive values for these array inputs.

src/components/submission-common.js
```diff
@@ -203,15 +203,16 @@ export const WithWarnings = props => {
...
   } = props
 
-  const [iconShape, iconColor] = message ? Utils.switchCase(message.type,
+  const [iconShape, iconColor] = Utils.switchCase(message.type,
     ['error', () => ['error-standard', colors.warning()]],
     ['info', () => ['info-circle', colors.accent()]],
-    ['success', () => ['success-standard', colors.success()]]
-  ) : [undefined, undefined]
+    ['success', () => ['success-standard', colors.success()]],
+    ['none', () => [undefined, undefined]]
```
cjllanwarne, line 210: What happens now if this _is_ undefined (or some other unexpected value)? Do we get a pageload error or does it show up as no content in the rendered page?

cjllanwarne (approved):

PR #176, opened 2023-06-07
[WM-1957] Struct 'Done' moves up one level

dillydally414:
Screen Recording 2023-06-07 at 3 03 47 PM

Clicking Done only moves you up one level if in an inner struct, instead of exiting out of the struct entirely.


mspector (approved):
Nice!

kpierre13 (approved):

PR #177, opened 2023-06-08
[WM-1984] Allow import method for either raw or original github WDL links

mspector:


src/libs/ajax.test.js
```diff
@@ -315,4 +315,45 @@ describe('Ajax tests', () => {
...
+      method_version: 'imported-version-4',
+      method_url: 'https://github.com/broadinstitute/warp/blob/develop/pipelines/skylab/scATAC/scATAC.wdl'
+    }
+    const body = JSON.stringify(payload)
+    const headers = { 'Content-Type': 'application/json' }
+
+    await cbasPact.addInteraction({
+      states: [
+        { description: 'ready to fetch myMethodVersion with UUID 90000000-0000-0000-0000-000000000009' },
+        { description: 'cromwell initialized' }
```
mspector, line 337: note that these states don't exist right now. this array should be updated as part of https://broadworkbench.atlassian.net/browse/WM-2010

mspector:
> how does storing the original github link vs raw make it easier to detect version control changes as described in WM-1981?

That's a good question -- I didn't even realize that detecting version control changes was a motivating factor for this feature. Let's bring it to tech talk.

> will this change make it so that the original (non-raw) link is always sent in place of a raw link, or does it only mean that the link that the user enters gets sent regardless of whether it is raw/original?

It only means that the link that the user enters will be sent regardless of the format

PR #179, opened 2023-06-12
[WM-2047] Use Default Outputs

dillydally414:
Updated recording (post-Chris' review) below

https://github.com/DataBiosphere/cbas-ui/assets/67511512/b5ab47ec-49f0-4387-a9d1-85631882597b


src/pages/SubmissionConfig.test.js
```diff
@@ -1566,6 +1566,79 @@ describe('SubmissionConfig inputs/outputs definitions', () => {
...
+            post: mockSearchResponse
+          },
+          types: {
+            get: mockTypesResponse
+          }
+        }
+      }
+    })
+
+    render(h(SubmissionConfig))
```
cjllanwarne, line 1596: TOL based on our conversation yesterday... are we testing the submission config page as a whole or the `OutputsTable` as a component? I would guess based on the scope of the changes that maybe these would be better as component unit tests?

src/pages/SubmissionConfig.test.js
```diff
@@ -2163,6 +2236,102 @@ describe('SubmissionConfig submitting a run set', () => {
       })
     )
   })
+
+  it('should call POST /run_sets endpoint with expected parameters after outputs are set to default', async () => {
```
cjllanwarne, line 2240: How does this kind of test separation sound? Do you think we'd cover our bases or would there still potentially be gaps?
- Does the SubmissionConfig page use the state set by the `OutputsTable` appropriately when POSTing (1 cross-component test)
- Does the `OutputsTable` set the state that's going to be used by the SubmissionConfig page appropriately? (many component-internal tests?)

cjllanwarne (approved):
LGTM, possibly a few test improvements to recommend/think about but they're not blockers

src/pages/SubmissionConfig.test.js
dillydally414: Yes, I agree this would be a good candidate for the separation we discussed yesterday. I wrote these tests prior to the discussion but it should be pretty easy to go in and shift them.

src/components/OutputsTable.js
```diff
@@ -27,6 +29,10 @@ const OutputsTable = props => {
     _.orderBy([({ [outputTableSort.field]: field }) => _.lowerCase(field)], [outputTableSort.direction])
   )(configuredOutputDefinition)
 
+  const setDefaultOutputs = () => {
+    setConfiguredOutputDefinition(_.map(output => _.set('destination', { type: 'record_update', record_attribute: _.last(output.output_name.split('.')) })(output))(configuredOutputDefinition))
```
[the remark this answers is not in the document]
dillydally414: I'll add this to Tuesday Tech Talk to see if Sara has any thoughts 👍

salonishah11 (approved):

PR #181, opened 2023-06-13
[WM-2008] View Workflow Script modal supports viewing Dockstore and Github workflows

salonishah11:
Tested locally:
Viewing workflow with Raw Github url source (this workflow is the pre-staged one)
Source: https://raw.githubusercontent.com/broadinstitute/viral-pipelines/v2.1.33.16/pipes/WDL/workflows/fetch_sra_to_bam.wdl

Viewing workflow with Github url source
Source: https://github.com/broadinstitute/viral-pipelines/blob/master/pipes/WDL/workflows/fetch_sra_to_bam.wdl

Viewing workflow with Dockstore source
Source: github.com/broadinstitute/viral-pipelines/fetch_sra_to_bam

Closes https://broadworkbench.atlassian.net/browse/WM-2008

### DIFF
--- a/public/config.json
+++ b/public/config.json
@@ -6,5 +6,5 @@
   "workspaceManagerRoot" : "",
   "workspaceId" : "",
   "containerResourceId" : "",
-  "dockstoreRootUrl" :  "https://staging.dockstore.org/search?descriptorType=WDL&entryType=workflows&searchMode=files"
+  "dockstoreRootUrl" :  "https://staging.dockstore.org"
 }
--- a/public/config.json
+++ b/public/config.json
@@ -6,5 +6,5 @@
   "workspaceManagerRoot" : "",
   "workspaceId" : "",
   "containerResourceId" : "",
-  "dockstoreRootUrl" :  "https://staging.dockstore.org"
+  "dockstoreRootUrl" : "https://staging.dockstore.org/"
 }
--- a/src/components/method-common.js
+++ b/src/components/method-common.js
@@ -32,7 +32,7 @@ export const submitMethod = async (signal, onDismiss, method) => {
 
 export const convertToRawUrl = (methodPath, methodVersion, methodSource) => {
   return Utils.cond(
-    // 3 Covid-19 workflows have 'Github' as source hence we check for that here to maintain backwards compatibility
+    // the case-insensitive check is to maintain backwards compatibility as 3 Covid-19 workflows have 'Github' as source
     [methodSource.toLowerCase() === MethodSource.GitHub.toLowerCase(), () => {
       // mapping of searchValues (key) and their replaceValue (value)
       const mapObj = {
--- a/src/components/method-common.js
+++ b/src/components/method-common.js
@@ -33,7 +33,7 @@ export const submitMethod = async (signal, onDismiss, method) => {
 export const convertToRawUrl = (methodPath, methodVersion, methodSource) => {
   return Utils.cond(
     // 3 Covid-19 workflows have 'Github' as source hence we check for that here to maintain backwards compatibility
-    [methodSource === MethodSource.GitHub || methodSource === 'Github', () => {
+    [methodSource.toLowerCase() === MethodSource.GitHub.toLowerCase(), () => {
       // mapping of searchValues (key) and their replaceValue (value)
       const mapObj = {
         github: 'raw.githubusercontent',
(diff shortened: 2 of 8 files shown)


src/pages/FindWorkflow/FindWorkflowModal.js
```diff
@@ -118,7 +118,7 @@ const FindWorkflowModal = ({ onDismiss }) => {
         ])
       ]),
       isSubHeaderActive('add-a-workflow-link') && h(ImportGithub, { setLoading, signal, onDismiss }),
-      isSubHeaderActive('go-to-dockstore') && div({ style: { marginLeft: '4rem', width: '50%' } }, [h(ButtonPrimary, { style: { width: 225 }, href: getConfig().dockstoreRootUrl }, ['Go to Dockstore'])]),
+      isSubHeaderActive('go-to-dockstore') && div({ style: { marginLeft: '4rem', width: '50%' } }, [h(ButtonPrimary, { style: { width: 225 }, href: `${getConfig().dockstoreRootUrl}/search?_type=workflow&descriptorType=WDL&searchMode=files` }, ['Go to Dockstore'])]),
```
salonishah11, line 121: I made this same as what we do in Terra UI.

src/pages/FindWorkflow/FindWorkflowModal.test.js
```diff
@@ -89,6 +89,6 @@ describe('FindWorkflowModal', () => {
     fireEvent.click(dockstoreSubHeader)
 
     const dockstoreButton = screen.getByText('Go to Dockstore')
-    expect(dockstoreButton).toHaveAttribute('href', 'https://staging.dockstore.org/search?descriptorType=WDL&entryType=workflows&searchMode=files')
+    expect(dockstoreButton).toHaveAttribute('href', 'https://staging.dockstore.org/search?_type=workflow&descriptorType=WDL&searchMode=files')
```
salonishah11, line 92: Not sure how this affects the opening of right Dockstore library, but I made this same as what we do in Terra UI.

public/config.json
```diff
@@ -6,5 +6,5 @@
   "workspaceManagerRoot" : "",
   "workspaceId" : "",
   "containerResourceId" : "",
-  "dockstoreRootUrl" :  "https://staging.dockstore.org/search?descriptorType=WDL&entryType=workflows&searchMode=files"
+  "dockstoreRootUrl" :  "https://staging.dockstore.org"
```
kpierre13, line 9: Thank you!

kpierre13 (approved):

src/components/method-common.js
```diff
@@ -24,18 +30,20 @@ export const submitMethod = async (signal, onDismiss, method) => {
...
-  } catch (error) {
-    notify('error', 'Error creating new method', { detail: (error instanceof Response ? error.text() : error) })
-    onDismiss()
-  }
-
-  return url
+export const convertToRawUrl = (methodPath, methodVersion, methodSource) => {
+  return Utils.cond(
+    // 3 Covid-19 workflows have 'Github' as source hence we check for that here to maintain backwards compatibility
+    [methodSource === MethodSource.GitHub || methodSource === 'Github', () => {
```
mspector, line 36: (I'm not adamant about this)

To avoid having to hardcode a particular capitalization of "github", I suggest making this check case-insensitive. The safest way to do that is probably something like:

`methodSource.toLowerCase() === MethodSource.GitHub.toLowerCase()`

src/components/method-common.js
```diff
@@ -24,18 +30,20 @@ export const submitMethod = async (signal, onDismiss, method) => {
...
+      // mapping of searchValues (key) and their replaceValue (value)
+      const mapObj = {
+        github: 'raw.githubusercontent',
+        'blob/': ''
+      }
+      return methodPath.replace(/\b(?:github|blob\/)\b/gi, matched => mapObj[matched])
+    }],
+    [methodSource === MethodSource.Dockstore, async () => await Ajax().Dockstore.getWorkflowSourceUrl(methodPath, methodVersion)],
```
mspector, line 44: you could probably include a case-insensitive check here, too, even though we don't know of any cases where case-sensitivity is a problem for dockstore workflows

src/pages/SubmissionConfig.js
```diff
@@ -224,7 +224,7 @@ export const SubmissionConfig = ({ methodId }) => {
   useEffect(() => {
     async function getWorkflowScript() {
       try {
-        const workflowUrlRaw = reconstructToRawUrl(selectedMethodVersion.url)
+        const workflowUrlRaw = await convertToRawUrl(selectedMethodVersion.url, selectedMethodVersion.name, method.source)
```
mspector, line 227: just curious... does this actually need `await`? It's not obvious to me why it would, but maybe I missed something

mspector (approved):
I left a couple of comments that I would consider optional / at your discretion

src/components/method-common.js
salonishah11: Yeah that's a good point. I will make the checking case-insentitive

src/pages/SubmissionConfig.js
salonishah11: I had to add await because we are making an Ajax call for Dockstore workflows

PR #183, opened 2023-06-15
[WM-1949] Searchable Inputs

dillydally414:
https://github.com/DataBiosphere/cbas-ui/assets/67511512/01c3a9d3-1b53-4db8-a04c-20a6ce10a834

### DIFF
--- a/src/components/InputsTable.js
+++ b/src/components/InputsTable.js
@@ -13,7 +13,6 @@ import {
   WithWarnings
 } from 'src/components/submission-common'
 import { FlexTable, HeaderCell, InputsButtonRow, Sortable, TextCell } from 'src/components/table'
-import { tableButtonRowStyle } from 'src/libs/style'
 import * as Utils from 'src/libs/utils'
 import { isInputOptional } from 'src/libs/utils'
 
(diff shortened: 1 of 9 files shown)


kpierre13 (approved):

src/components/InputsTable.js
```diff
@@ -104,94 +107,97 @@ const InputsTable = props => {
...
+    h(InputsButtonRow, {
+      optionalButtonProps: {
+        includeOptionalInputs, setIncludeOptionalInputs
+      },
+      setFromDataTableButtonProps: {
+        inputRowsInDataTable, setConfiguredInputDefinition
+      },
+      searchProps: {
+        searchFilter, setSearchFilter
+      }
```
salonishah11, line 120: It seems Terra UI has a component called `DelayedSearchInput` for filtering inputs. Maybe we can use that here so that we can reuse components especially since we are reintegrating back into Terra UI. One example of this component is in The `Inputs` tab in the Submission Config page in a GCP workspace:
Screenshot 2023-06-27 at 5 03 31 PM

dillydally414: Yes, the component used here was copied over from the Terra-UI version so it should be pretty easy to merge back with Terra

salonishah11: Ah I see. Apologies that I missed it.

salonishah11 (approved):

PR #187, opened 2023-06-22
[WM-1961] Typecheck WDS Lookups

dillydally414:
https://github.com/DataBiosphere/cbas-ui/assets/67511512/944b4f5e-d293-419e-90ed-e00e6c8a0447

Record lookups are only allowed to be selected from WDS attributes that match expected input types (or attributes that we can coerce, e.g. Int column provided to String input). Record lookups are also validated by this type matching, in case the user switches tables and now has records selected that exist in the new table but have a different type.

### DIFF
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -380,6 +380,7 @@ export const typeMatch = (cbasType, wdsType) => {
       ['Int', () => wdsType === 'NUMBER'],
       ['Float', () => wdsType === 'NUMBER'],
       ['Boolean', () => wdsType === 'BOOLEAN'],
+      ['File', () => wdsType === 'FILE' || wdsType === 'STRING'],
       [Utils.DEFAULT, () => true]
     )
   } else if (unwrappedCbasType.type === 'array') {
--- a/src/components/submission-common.js
+++ b/src/components/submission-common.js
@@ -373,18 +373,29 @@ const validateRequiredHasSource = (inputSource, inputType) => {
   } else return { type: 'error', message: 'This attribute is required' }
 }
 
-const validateRecordLookup = (source, recordAttributes) => {
-  if (source) {
-    if (source.type === 'record_lookup' && !recordAttributes.includes(source.record_attribute)) {
-      return { type: 'error', message: 'This attribute doesn\'t exist in the data table' }
-    }
-    if (source.type === 'object_builder') {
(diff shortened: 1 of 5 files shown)


kpierre13 (approved):

src/components/submission-common.js
```diff
@@ -373,18 +373,34 @@ const validateRequiredHasSource = (inputSource, inputType) => {
...
 
-const validateRecordLookup = (source, recordAttributes) => {
-  if (source) {
-    if (source.type === 'record_lookup' && !recordAttributes.includes(source.record_attribute)) {
-      return { type: 'error', message: 'This attribute doesn\'t exist in the data table' }
-    }
-    if (source.type === 'object_builder') {
-      if (source.fields) {
-        const fieldsValidated = _.map(field => field && validateRecordLookup(field.source, recordAttributes) === true, source.fields)
+export const typeMatch = (cbasType, wdsType) => {
```
salonishah11, line 376: Does this need cases for type `File` and `JSON`? I think WDS does have `FILE` and `JSON` data types while CBAS has `File` and `Object` types as data types

dillydally414: I will look into this, and it's worth noting that any cases not explicitly handled will pass the comparison (so not including cases will err on the side of allowing the input instead of rejecting the input). Allowing any type input for File would be consistent with our Type a Value validation for that type (`isPrimitiveTypeInputValid`).

salonishah11: Any reason not to include the `Struct` type in the check?

dillydally414: This check will be performed fieldwise on Structs that use the object builder and have record lookups for their fields. I wasn't sure if there's a way to easily see what fields a WDS `JSON` has and cross-check with the struct fields CBAS expects, but I suppose it's better to restrict Struct lookups to only JSON objects rather than anything (even if we can't check further)

salonishah11: Yeah I think we for Struct inputs and `Fetch from Data table` as source, we can ensure that the attribute has `JSON` type and then let CBAS handle if the fields in the JSON object are actually what Struct input needs. What do you think?

dillydally414: Addressed in latest commit.

salonishah11 (approved):

PR #188, opened 2023-06-27
[WM-2055] Remove workflow version input box

kpierre13:


src/components/ImportGithub.test.js
```diff
@@ -56,14 +54,12 @@ describe('Add a Workflow Link', () => {
 
     const urlLink = screen.getByPlaceholderText('Paste Github link')
     const workflowName = screen.getByPlaceholderText('Workflow Name')
-    const workflowVersion = screen.getByPlaceholderText('Workflow Version')
     const addToWorkspaceButtonDisabled = screen.getByLabelText('Add to Workspace button')
 
     expect(addToWorkspaceButtonDisabled.getAttribute('aria-disabled')).toBe('true')
 
     fireEvent.change(urlLink, { target: { value: githubLink } })
```
dillydally414, line 61: Maybe add an `expect` here for the name autofill prior to changing it?

dillydally414 (approved):

salonishah11:
@kpierre13 it looks like this branch is inheriting changes from some other branch or the main branch? Maybe it needs to merge the changes from main to this branch?

src/components/method-common.js
```diff
@@ -47,3 +47,14 @@ export const convertToRawUrl = (methodPath, methodVersion, methodSource) => {
     }
   )
 }
+
+export const getMethodVersionName = url => {
```
salonishah11, line 51: Is it possible to get the version tag from Github itself using the Github url?

kpierre13: That should be possible, and the `else` statement here should cover getting the version tag from a github.com link. Are you referencing a different kind of github url?

salonishah11: Sorry waht I meant was ping Github or its APIs to obtain the version or branch name using the Github url itself rather than doing the String manipulation.

kpierre13: Ah yes, I tried to look for an API that could get this for us, but it seemed for all of the potential endpoints, I needed the branch name anyway. I'll give the documentation a once over one more time for good measure.

salonishah11: Oh I see. And just to make sure, the position of version tag / branch name is always consistent? Hence there is no risk with doing string splicing this way?

kpierre13: Not that I was able to find. I tried methods from different repositories, as well as trying files at different depths in their respective directory trees.

salonishah11: I see. I guess it might be worth adding different url tests (apart from the 3 you already have) for this method to ensure we don't extract the wrong version / branch name for the url.

salonishah11 (approved):
small nit about adding more url tests but otherwise it LGTM

PR #189, opened 2023-06-28
[WM-2050] flaky tests

kpierre13:


dillydally414 (changes requested):
I seem to still be getting test failures occasionally locally for each of the tests outlined in the ticket (`'should display warning for required inputs for a newly imported method'`, `'should display warning icon/message at each level of the struct builder when a field has a missing attribute'`, `'should call POST /run_sets endpoint with expected parameters after struct has been updated'`).
If it helps, here's how I reproduce the flakiness without hitting run 5 million times:
`it('<Test Name>', ...)` -> `it.only.each(Array(100).fill(1))('<Test Name>', ...)`
This will run the specified tests 100 times - you can also speed it up slightly by using `.only.concurrent.each` but this is experimental so not sure if it has an impact on test failure rate.

dillydally414 (approved):
🥳  if only we took the test confidence survey this week

PR #201, opened 2023-09-06
[WM-2183] Remove auto-updating CBAS UI image in cromwhlem charts

salonishah11:
JIRA ticket: https://broadworkbench.atlassian.net/browse/WM-2183

### DIFF
--- a/.github/workflows/tag-and-publish.yml
+++ b/.github/workflows/tag-and-publish.yml
@@ -71,33 +71,6 @@ jobs:
       - name: Push GCR image
         run: docker push ${{ steps.image-name.outputs.GCR_NAME }}
 
-      - name: Clone Cromwhelm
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-and-publish.yml
```diff
@@ -71,33 +71,6 @@ jobs:
       - name: Push GCR image
```
salonishah11, line 71: I have left publishing of new image purposefully since there is no harm in having it here and only removed the updating helm charts portion from the Github workflow.

mspector: and this whole repo is going away soon, anyway!

mspector (approved):

cjllanwarne (approved):

